PR #1496, opened 2015-01-28
Icon for Translator. Fixes #1465

palob:
Icon for Translator, issue #1465
Pushed:
translator
Alternative background:
translator1

PR #1504, opened 2015-01-28
Translator icon - colourful bubbles

palob:
Now in colourful colours :rainbow:

translator-new

See #1496 for the old one.

### DIFF
--- a/Numix-Circle/48x48/apps/translator.svg
+++ b/Numix-Circle/48x48/apps/translator.svg
@@ -1,0 +1,322 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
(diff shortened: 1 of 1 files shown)


Foggalong:
Check your pull, can't be automerged because of clashes

palob:
Yeah, that's since my base fork is without the old icon.

palob:
You could revert #1496, this should make automerge possible.

Foggalong:
I wasn't sure about the colours on this so asked Numix folk. Consensus is that the colours are fine, just the purple needs to be a bit less dominant

palob:
Like this?
translator

ChuckDaniels87:
Really nice ;)

Foggalong:
Yeah, that's the one. Push it and I'll sort out the merging

palob:
Pushed :+1:

PR #1513, opened 2015-01-29
Enhanced icon for Muon Discover

palob:
The current icon is the generic package manager icon. But since it is somethin like GNOME Software, I pushed this icon:
muondiscover
(+ changed symlink, added extra symlink)

Alt:
muondiscover2


Foggalong:
Thanks for this! I will at some point have to update all the software icons to the lastest design so I'll do that then

PR #1561, opened 2015-02-07
Added a tray icon for nvidia

alfredvaa:
nvidia
I added a grey version of the nvidia icon to match the other grey icons in Unity tray (upper right corner).


alfredvaa:
nvidia
Changed color to this instead.

Foggalong:
For status icons we don't stick to the circle design, so it would just be the Nvidia logo. Take a look at the SVGs of our other icons for the colours needed :)

alfredvaa:
nvidia
I see! I have updated it now!

Foggalong:
Almost spot on, just need to add the undershadow as shown here
screenshot - 080215 - 20 44 39

alfredvaa:
That should be it!

PR #1577, opened 2015-02-15
Added Delicious (delicious.com) icon

Atema:
web-delicious.svg


palob:
I see the current colours are the original ones but maybe using a shade of gray instead of black (black hides the bevel) + desaturating the blue part would make the icon more consistent with the exiting ones.

Foggalong:
I'm gonna merge this in and then fix since it seems OP doesn't want to

PR #1706, opened 2015-03-12
Circular Games app category icon

palob:
Added a Numix Circle icon for the Games app category
applications-games


Foggalong:
We still haven't decided if we want Circular icons for categories which why it's all still inconsistent. Hold the fort on this one :)

palob:
Alright, here is an alternative which probably scales better:
applications-games1

palob:
Deferred.

Foggalong:
@palob fancy submitting the first of those as a solution to [this](https://github.com/numixproject/numix-icon-theme-circle/issues/1688)?

PR #1729, opened 2015-03-16
correct icon name for StackEdit

m-e-h:


Foggalong:
The icon was reported with the other name so I'm assuming both are needed. Can you make this a symlink instead?

m-e-h:
OK changed to a symlink. That was a weird name for a chrome app. :smiley:

Foggalong:
Yeah, looked weird to me too :P Thanks for adding the normal Chrome link too though :)

PR #1734, opened 2015-03-17
Fixed Tomboy icon

palob:
Before:
tomboy-old
After:
tomboy


Foggalong:
Good catch!

ChuckDaniels87:
Nice one!

PR #1753, opened 2015-03-20
Scribus icon fixes

palob:
Fixed this
scribus-fix
+alignment


palob:
This time Github calls it invalid but Firefox displays it correctly.

PR #1780, opened 2015-03-26
Added Spark IM Client icon

Ghokun:
This is an icon for the Spark IM Client. Original icon is hardcoded so one has to create .desktop entry to use this icon.

Here is an example:
[Desktop Entry]
Name=Spark
Version=2.6.3
GenericName=Spark
X-GNOME-FullName=Spark
Comment=ignite realtime Spark IM client
Type=Application
Categories=Application;Utility;
Path=/opt/spark
Exec=/bin/bash Spark
Terminal=false
StartupNotify=true
Icon=spark
TargetEnvironment=Unity

The original icon was:

spark

Proposed icon:

spark

As a side note tray icons are also hardcoded in the spark.jar file, so I have created this repo for tray icons. https://github.com/Ghokun/spark-numix-tray-icons


Ghokun:
I have a little problem to mention:

problem

Icon looks fine in every place but dash. Do I have to make a change for that?

Foggalong:
Nice start, but try and removed the gradients and outlines - keeps it more in line with the rest of the icons :)

Ghokun:
Removed gradients, i think it suits better now.

spark

Foggalong:
Looks great :D

PR #1788, opened 2015-03-26
Create steam_icon_217690.svg

hlechner:
I've changed the color to pure black #000000 and also I've applied the 25 to alpha, returning this into RGBA: 00000025 (I made it wrong?)

To fix the pixels I've changed +1 X and -1 Y respective, I hope it reflect in same amount of pixels


Foggalong:
Looks right to me :)

PR #1936, opened 2015-04-11
Added velocibox icon

Atema:


palob:
Nice to see new contributors. 
Two issues with your design: 
+ Vertical/horizontal lines (E, L and I) aren't aligned to the pixel grid which makes them look blurry
+ The baseplate gradient is to intense (10L lighter at the top is what the guidelines say)

Atema:
Thank you for your comments, I will try to solve those issues as soon as possible.
By the way, have you got a link to those guidelines. I was looking for the gradient rules, but couldn't find any documentation.

palob:
[Numix Circle guidelines](https://github.com/numixproject/numix-wiki/wiki/Guidelines)

Atema:
I've changed the background gradient and tried to fix the alignment as much as possible.

palob:
Still a mixture of sharp and blurry lines. Currently all letters take up max 8px X 8px. How about going for 7x7 for the letters? This way you could align the vertical *I* line and the horizontal *E* line.

Foggalong:
^ 7x7 would be a good idea. And yay new contributors :dancer:

Atema:
I've decreased the E and L icon size to 7x7, and the I to 7x8; a bit larger, but I guess that's the only way ;(

palob:
Looking at the original icon even 6x8 for the I might be possible.
Now there's the issue left that the letters without horizontal/vertical lines are darker apparently when the icon is viewed at 48x48. Makes the E, L and I stand out.
This could be counteracted by darker E, L and I letters (which however will show when the icon is viewed at bigger sized).

Or by departing from the original design and redoing the letters as pixel art.

Atema:
I'll take a look at it tomorrow and see what I can do

palob:
@Atema Now that you closed this I hope my criticism wasn't discouraging.
You're not far off :)

Atema:
I didn't mean to close it; I just cleaned my repo, so at the end there
aren't tens of new commits in it ;) I'll update this when I've got some new
icons :)

Issue #1953, opened 2015-04-12
Icon for Pixsta (Google Chrome App)


ahlee2326:
https://chrome.google.com/webstore/detail/pixsta/cijncchffkmlnfdbnkkfclcbnjcoegjc?utm_source=chrome-app-launcher-info-dialog

bil-elmoussaoui:
@Foggalong Can i symlink this to the default instagram app icon ?

PR #1954, opened 2015-04-12
Symlink Pixsta to unity-webapps-instagram

bil-elmoussaoui:
Fix this #1953


Foggalong:
Thought it was distinct enough to warrant it's own icon, so put one together in 76821cc :)

PR #2005, opened 2015-04-18
Fixes #1907

### LINKED ISSUE #1907: Icon for Quake
classic FPS from id Software
icon = quake

icon as supplied by id Software, converted to png
quake

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10749524-icon-for-quake?utm_campaign=plugin&utm_content=tracker%2F514195&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F514195&utm_medium=issues&utm_source=github).
</bountysource-plugin>

wa4557:
Easy one. Symlink to other quake issues as well or will they get distinct icons?
quake


Foggalong:
Yeah, all the different Quakes have slightly different symbols which I think vaguely relate to the roman numerals of the version number, so a unique icon for each would be good. We already have on for IV too I think :)

Foggalong:
Sorry, just checked and Quake 3 is the one we have.

wa4557:
Fine, I added quake 1,2 and 4 used always the same background but different symbols

PR #2083, opened 2015-05-08
New Audacity icon

dirtydancing:
Fixes https://github.com/numixproject/numix-icon-theme-circle/issues/1158. Cf. also conversation over there.

audacity


palob:
How does it look with less saturated colours and maybe a darker blue on the headphones for better contrast?

dirtydancing:
Good call! A little less saturation (not too much less esp. for the yellow and red, as otherwise the icon will lose its character) and a somewhat darker blue. Easier on the eyes, and it also scales better. This might seem like a nuance, but I think it makes for quite an improvement. Pushed :)

audacity

comparison

palob:
Alright, let's see if it's ready ro go.
In terms of vividness you can counteract reduced saturation with increased lightness to a certain extent sometimes.

dirtydancing:
Corrected alpha channel for yellow, I just noticed that for some reason I must have accidentally changed that  when changing other values. Still the same icon, though :)

audacity

PR #2106, opened 2015-05-15
Icon for MKVToolNix on Arch Linux

bernimoses:
There is a new icon name in the package desktop file. I added a symlink.


Foggalong:
Thanks!

PR #2107, opened 2015-05-15
Added icon for Pogo

dirtydancing:
Fixes https://github.com/numixproject/numix-icon-theme-circle/issues/2076.

pogo


Foggalong:
For anyone reading this, icons like this are the exception rather than the rule for our icon design guidelines. I'm merging this one because it looks pretty great, sticks to the original icon, and isn't far enough from the guidelines that it'll sit out like a sore thumb.

Issue #1930, opened 2015-04-11
MintInstall missing icons.


spelufo:
![mintinstall-numix-circle](https://cloud.githubusercontent.com/assets/1312203/7101511/3d7c0218-e033-11e4-82dc-23f3f1934fe2.png)

I don't know why, but mintinstall is not taking some of numix-circle icons. I think this might be fixed with some symlinking, since most of the categories can be replaced with existing icons. The needed icons are named `applications-*`, as can be seen [here](https://github.com/linuxmint/mintinstall/blob/6ffef49e98ca8aa27f8227b14a492476ea700bbe/usr/lib/linuxmint/mintInstall/mintinstall.py#L1117-L1183). I tried [this](https://github.com/numixproject/numix-icon-theme-circle/pull/1678) but it didn't work. I guess they have to be in "categories" subdirectories. Regular Numix has most of them: try `find /usr/share/icons/Numix/ -name 'applications-*.svg' | cut -d '/' -f 6-`.

Also, how can I clone the repo and install my local version of it with apt, do I have to package it? How do you do it?

Kat1031:
What's your window borders and desktop theme? Looks good.

Foggalong:
Are you running the latest version of the theme from git?

bil-elmoussaoui:
If you want to update the icons to the last update you can use this : 

  ```bash
git clone https://github.com/numixproject/numix-icon-theme.git
git clone https://github.com/numixproject/numix-icon-theme-circle.git

cp -r ~/numix-icon-theme/Numix /usr/share/icons/
cp -r ~/numix-icon-theme/Numix-Light /usr/share/icons/
cp -r ~/numix-icon-theme-circle/Numix-Circle /usr/share/icons/
cp -r ~/numix-icon-theme-circle/Numix-Circle-Light /usr/share/icons/

gtk-update-icon-cache /usr/share/icons/Numix-Circle/
gtk-update-icon-cache /usr/share/icons/Numix-Circle-Light/
  ```
To avoid issues , as the monochrome icons are moved to the base theme. You better remove the old version you use using this 
  ```bash
rm -rf /usr/share/icons/Numix-Circle/
rm -rf /usr/share/icons/Numix-Circle-Light/
rm -rf /usr/share/icons/Numix/
rm -rf /usr/share/icons/Numix-Light/
  ```
You need to to this using the root permissions.

palob:
It is at least a revision where #1678 is already applied.

[...]

PR #2172, opened 2015-05-25
Removes circular icons that are also in base

### LINKED ISSUE #2164: Cinnamon Menu Consistency Issue
numixgitissue

Issue is illustrated in above image.

I would give an example symlink but there doesn't seem to be a Numix Circle equivalent for that controller icon, I do however know that the name of that icon is "applications-games". This also doesn't seem to be the only "applications-" prefix icon that is missing an icon or symlink either, fixing this would probably prevent further consistency issues.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/17120143-cinnamon-menu-consistency-issue?utm_campaign=plugin&utm_content=tracker%2F514195&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F514195&utm_medium=issues&utm_source=github).
</bountysource-plugin>

wa4557:
Fixes #2164 and #1930


bil-elmoussaoui:
You forgot ``` applications-php.svg" this icon is used for
``` php.svg 
phpmyadmin.svg```

wa4557:
Yeah that was intenional because there's no corresponding icon in base

bil-elmoussaoui:
the best thing is to remove it and report it in base? as this icon is only used in Software center (i guess)

wa4557:
Yeah I'll probably do this in a pull request afterwards. I don't want to put too much effort in creating a php categories icons, if this PR isn't merged

palob:
This might have been discussed before but are ``help.svg`` and ``info.svg`` actually app icons?
Are they used for any system settings plug?

Problem is, they are shown in the interfaces of some programs.

Foggalong:
Gonna close this because these kind of things are gonna have to be considered on a per app basis. Hardcoded stuff is complicated

wa4557:
Makes kind of sense but concerning #2164 and #1930 what's then the preferred way to fix it? Adding the missing icons to circle or removing those icons from circle (then this PR would have made sense after all)

Foggalong:
I honestly have no idea. Whenever the topic has come up within Numix we've been unable to agree. Maybe it's something we should ask G+ users?

wa4557:
Sounds like a plan :+1:

PR #2186, opened 2015-05-28
Add icon for Reportbug

gregorydk:
Made debian-swirl.svg which is used by Reportbug (https://wiki.debian.org/reportbug)


Foggalong:
We already have `distributor-logo-debian.svg` so the icon just needs to be a symlink to that

gregorydk:
I know, but instead of ending up with 10 identical icons, I wanted some variation, so I inverted the colours, which also makes it resemble the real icon/logo more.

Foggalong:
It's a nice idea but not really necessary for two very niche icons

gregorydk:
It's fine. You can make it a symlink then.

gregorydk:
Do I need to open a new issue for this?

Foggalong:
Not at all, [already been fixed](https://github.com/numixproject/numix-icon-theme-circle/commit/e061730f90851a4df774cc2004a63c52fe9e1f78) :)

gregorydk:
Awesome :-)

PR #2238, opened 2015-06-01
Icon for Wolfenstein: The New Order

Eleuin:
Yet another icon for this series, this leaves just Wolfenstein 3D left


palob:
Compared to existing icons the colours look brash to me. Try a less saturated (lower S value in the HSL notation) and maybe brighter baseplate and use white L249 instead of pure white.

Eleuin:
there we go
wolftno

palob:
Still a bit dark and saturated but let's see what's @Foggalong's view

Eleuin:
ill resubmit this PR, just had a commit i wanted to whack because i don't need silly commits clogging up the repo :>

Issue #2016, opened 2015-04-21
KDE breakage


palob:
Despite having done a few KDE icons lately I have never actually used KDE as my DE.

Yesterday I tried Numix Circle on *Kubuntu 15.04* and *Fedora 21 KDE* live isos and in addition to a few missing icons in the base theme a lot of Numix Circle icons appear broken all over the place (Menu, Dolphin, Gwenview, Muon Discover...).

It's not the breakage encountered on Github or in web browsers but looks like this:
![breakage](https://cloud.githubusercontent.com/assets/7050624/7252079/bc7cfe86-e831-11e4-94b7-28e66aea420b.png)

So the issue is about paths exceeding the baseplate, recoloured objects showing the colour set previously and recoloured baseplates being transparent. The latter two might be somehow related two our dreaded Inkscape gradient bug.

It looks the same with other Qt-based viewers (like Nomacs).
The issue is due to KDE/Qt using a different SVG rendering backend system (Qt SVG instead of librsvg).

[According to this](https://forum.kde.org/viewtopic.php?f=19&t=121637) the KDE bugtrackers are full of related reports but the problem is with Qt and not likely to be fixed anytime soon.

Some more reports of Qt's mediocre SVG support:
http://www.qtcentre.org/threads/57772-How-to-replace-Qt-svg-render-engine
https://bugreports.qt.io/browse/QTBUG-12588
https://bugs.kde.org/show_bug.cgi?id=333497
https://bugreports.qt.io/browse/QTBUG-1865
https://bugreports.qt.io/browse/QTBUG-28636
http://matplotlib.1069221.n5.nabble.com/Qt-SVG-clipping-bug-it-s-NOT-a-matplotlib-bug-Qt-won-t-fix-td37980.html

Not sure if anything can/should be done on the Numix side, just writing this so as to anticipate similar bug reports.

palob:
If need be a script rendering the SVGs to PNGs for all sizes (using a tool capable of doing this reliably) could be a workaround.

Foggalong:
I think PNGing stuff is out of the question - you loose out on a lot of the quality in scaling it. It'd make sense for the theme that has different icons for different sizes (like Moka) but not for Numix.

It's possibly a case of going through all the icons and making sure they use intersections rather than path clipping. I'm not sure how to deal with the gradient & vanishing baseplate objects though :(

palob:
[...]

PR #2240, opened 2015-06-02
SVG cleanup + KDE/Qt fixes

palob:
Did some SVG cleanup by removing invisible shadow elements, path unions of shadow elements and executing [your SVG cleaning script](https://github.com/numixproject/numix-tools/blob/master/numix-tools/svgclean) (found it, finally :) + fixed numerous small issues (like broken strokes, stroke -> path conversions, miscoloured shadows).

A welcome side effect is that it fixes the baseplate part of the KDE/Qt issues #2016.

Total file size reduction: 4.1 MB.


palob:
Forgot some, will add them later.

Foggalong:
The reason we stopped using our own cleaner was because the reduction to a single line started causing problems with elements in icons disappearing. Is that not a problem anymore?

palob:
All pre-mid 2014 were in the single-line state anyway. I tested with several distros and didn't come accross missing elements. 
Maybe it was about elements missing in the GitHub preview?

Foggalong:
No, it was actual missing elements from the icons. One I remember was the Rhythmbox speaker kept vanishing. It's the reason we stopped cleaning icons in that way. You sure that's not happened here?

palob:
Maybe it was some grouping issue? I'll do some even more intense testing :)

palob:
Do you remember when exactly this vanishing did occur?

Foggalong:
It was when we last tried using it, I think early this year

palob:
So I went through the icons over and over again (using different distros / file managers / image viewers) but I couldn't make out any defective icon.

wa4557:
@palob I'll try to verify this too :)

palob:
@wa4557 Thanks a lot :)

bil-elmoussaoui:
I verified this too using elementary OS freya and everything works fine! 
capture du 2015-06-03 12 45 32

wa4557:
I looked at all the icons. No problem when using the new icons at Ubuntu 15.04 + unity

dirtydancing:
I checked with Arch/GNOME Shell: no visible breakage with the new icons here either.

However, I noticed one thing about Inkscape: I checked with `unsettings.svg`, an icon that has caused some problems with the gradient before, cf. https://github.com/numixproject/numix-icon-theme-circle/pull/2052. When opening with Inkscape, with the current icon I was able to easily modify the gradient colour. In contrast, with the new icon, for some reason I could not modify the gradient colour at all. Now, this might be just this icon, or this might be just me using Inkscape, but maybe someone can reproduce this effect?

In addition, as the metadata etc. are gone in the cleaned up icons, in Inkscape they do not open maximised or with the grid visible anymore, but I figure this is intended.

dirtydancing:
One more aspect for your consideration: to me, when opening the new .svg files with a text editor, the code does not necessarily look "cleaner", but rather more "compressed" (into one line). This would make troubleshooting details within the code itself more difficult.

I am just wondering (not opposed to this change): Is there a way to "decompress" the code again? And is there any advantage to the single line compression besides reducing the file size?

palob:
Thanks to all for your testing.
@dirtydancing I can reproduce the *unsettings.svg* thing. I have to set the colour to solid and to gradient again in order to edit it. Don't know what that's about.

Unravelling the single-line SVG should be fairly trivial, it's XML and each line starts and ends with **<**/**> respectively. Of course after opening an SVG in Inkscape and saving it, you get back to the uncopressed variant, losing all optimisations.
I don't know which option in the SVG cleaning script exactly causes the one-line files, maybe it's possible without that one.

palob:
``svgcleaner-cli --help`` shows *Save file with only required whitespace and newlines* for ``--compact-output``

dirtydancing:
Confirmed that the .svg code turns back to "uncompressed" after saving an icon file with Inkscape.

I would agree that it makes sense to use the cleaner script in a way that preserves the separate lines.

For some reason, I am still stuck with this "bug" that lets me change the gradient colour visibly in Inkscpape, but not visibly e.g. in Nautilus or Thunar. So this is not limited to Ubuntu, but also appears in Arch. Somehow, there is a "doubling" of the stop-color/opacity for the gradients when I change the gradient colour. Well, I figure as @Foggalong has already said earlier, to not get bogged down by this.

Screenshot of the code
vlc_gradient

VLC (vlc2.svg) icon change in Inkscape (different colours)
vlc2_inkscape

versus e.g. Nautilus (no difference displayed for vlc2.svg)
vlc2_nautilus

But this thing apparently is not related to the present changes, so I guess this is safe to ignore here :)

palob:
Yeah, we had to deal with this "gradient bug" earlier.

Foggalong:
@palob RE your comment in the other merge thread, I didn't think this was mergeable yet because of the gradient bug? Feel free to correct me if otherwise!

palob:
Sorry, which thread exactly are you referring to?

dirtydancing:
@palob I believe @Foggalong is referring to this comment of yours: https://github.com/numixproject/numix-icon-theme-circle/pull/2247#issuecomment-109418687

@Foggalong I believe the above mentioned "gradient bug" has nothing to do with the clean-up filed here by @palob. This "bug" has appeared before, and this clean-up as far as I can tell has no influence on that (neither for the better nor for the worse). imo this pull request here is ready to be merged.

Foggalong:
@palob @dirtydancing I thought you were discussion a problem about the usettings gradient being corrupted, caused by this cleanup?

palob:
We were indeed. The gradient wasn't corrupted in terms of the outer apperance. The glitch was with editing.
*unsettings.svg* had been a strange one before with big filesize despite the simple shape and colours. I've copied the symbol to anpther baseplate and now I can edit the gradient without the need to set a solid colour first + the filesize is halved.

palob:
Regarding the single-line SVGs: You can restore the multiline layout (while losing the SCG cleanups) in an easy way by batch-running any Inkscape-CLI option (for instance ``inkscape --vacuum-defs``).

If desired I could do this and run the SVG cleaning script afterwards. This way we'd get cleaned and Qt-fixed files while retaining readability. (I'd go through the icons again afterwards :).

Foggalong:
@palob I'd prefer multiline readout yeah :) Thanks for looking at this, it's very helpful!

palob:
Alright, I'll probably do it as a new PR later.

palob:
Yay, I've found a way to fix the "gradient bug" for good on all icons. I'll do it tomorrow and fix the ``svgclean`` script accordingly.

dirtydancing:
Great! :+1:

Foggalong:
@palob feel free to submit a pull request to the numix-tools repo with the fix :D

palob:
A little explanation:

In a SVG with a fresh baseplate gradient (no gradient bug) a baseplate gradient looks like this:

```SVG
    <linearGradient
       id="linearGradient3794">
      <stop
         style="stop-color:#5c431c;stop-opacity:1;"
         offset="0"
         id="stop3796" />
      <stop
         style="stop-color:#6b4e21;stop-opacity:1;"
         offset="1"
         id="stop3798" />
    </linearGradient>
```

Both stop colours and opacities are defined as ``style``.

After applying the current unfixed *svgclean* script it would look like this:
```SVG
  <linearGradient id="linearGradient3800" y1="28" x2="0" y2="20" gradientUnits="userSpaceOnUse">
   <stop stop-color="#5c431c" stop-opacity="1"/>
   <stop offset="1" stop-color="#6b4e21" stop-opacity="1"/>
  </linearGradient>
```
Here stop colours and opacities are defined seperately as ``stop-color`` and ``stop-opacity``.
Now when editing the gradient colours in Inkscape the changes are saved as ``style`` after the values for ``stop-color`` and ``stop-opacity`` which are left untouched.
The "gradient bug" occurs because the latter two values take precedence for what's displayed in the file manager / image viewer / web browser.

The simple fix:
The ``svgcleaner-cli`` option ``--join-style-atts`` merges the ``stop-color`` and ``stop-opacity`` values into ``style`` attributes.

palob:
I'll apply the fix in a new PR

PR #2243, opened 2015-06-03
Icon for pumpa

gregorydk:
Icon for pumpa (https://pumpa.branchable.com/), a pump.io client.

### DIFF
--- a/Numix-Circle/48x48/apps/pumpa.svg
+++ b/Numix-Circle/48x48/apps/pumpa.svg
@@ -1,881 +1,127 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
-   xmlns:cc="http://creativecommons.org/ns#"
-   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
-   xmlns:svg="http://www.w3.org/2000/svg"
-   xmlns="http://www.w3.org/2000/svg"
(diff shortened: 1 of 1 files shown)


palob:
The baseplate colour is quite harsh and dark, isn't it?

gregorydk:
If I make it lighter there's not enough contrast with the actual icon, unless I use light grey. But I don't want to make yet another icon with a grey baseplate. Any suggestions? Green baseplate?

pumpa-green

palob:
I think the contrast looks OK on the revised one + the grid doesn't look that aggressive.
Maybe even a tad lighter / more yellowish.

A minor issue: there are a few unneeded nodes on the shadow.

gregorydk:
I don't see the unneeded nodes.

New:
pumpa

palob:
Yep, looks good to me :+1:. What do you think?
I was referring to the nodes between the edges :)

Foggalong:
Icons that use pixel art need 100% aligning to grid. They'll just end up looking like rubbish otherwise. The symbol is a bit too big too. We recommend 24px<sup>2</sup> for symbol size but this one is 30px by 25px. Reduction to about 26px by 22px would make it fit the guidelines a bit better

gregorydk:
The current size is 12 x 10 "pixels". So 24 x 20 fits better.

Foggalong:
24x10 would be acceptable, though it'll look a little small

gregorydk:
If I make it 26 px wide, the height is 21.667 px. Not really ideal I think. So 24 x 20 or 30 x 25 are round numbers.

Foggalong:
24px would be the ideal, but remember the rest of the lines have to be aligned to grid too

PR #2247, opened 2015-06-05
Fixed icon breakage

Eleuin:
Exorcised the daemons (random shape in the DeviantArt icon)


Eleuin:
@Foggalong no idea how this wasn't noticed

Foggalong:
What random shape am I looking for?
screenshot_2015-06-05_17-10-59


But anyway, this has brought to my attention that the dA icon needs major fixing. The circle part of the symbol needs converting from stroke to path and then the shadow needs redoing so that it's black and doesn't have the weird circle thing.

Eleuin:
fuglymess
this ugly mess, i think my OCD is compelling me to fix this, despite it not really being my typical icon (i have a few others in the works that i have more final concepts, but i think this is more important), this really does need fixing

Foggalong:
That shape as far as I know is supposed to be there. It's for [this Chrome app](https://chrome.google.com/webstore/detail/deviantart-muro/namljbfbglehfnlonjmebceimaalofei) which isn't the normal dA. It's been this way [for years now](https://github.com/numixproject/numix-icon-theme-circle/blob/758ea8c52f918f56a740034b9169adaf8d9c3976/Numix-Circle/48x48/apps/chrome-namljbfbglehfnlonjmebceimaalofei-Default.svg) (before I joined Numix, this was just the earliest commit I could find) so I'm guessing that's what the original icons looked like

Eleuin:
it really should look more like this (which although its not the current dA logo, it is the one used by the icon)
da

Eleuin:
so obviously something like this looks a bunch better (i converted the strokes to paths, redid the shadow, and merged the two shapes that ARE supposed to be there)
daicon

Eleuin:
there we go, that should honestly be a bunch better

Eleuin:
but seriously, that giant blob has had enough innuendos made on IRC that it deserves to be purged and then have the original posted to /r/unixporn

Foggalong:
If you wish to make the two icons the same then link the two together - `ln -s web-deviantart.svg chrome-namljbfbglehfnlonjmebceimaalofei-Default.svg`

palob:
Just saying, I've fixed all those issues in the SVG cleaning PR  :)

Eleuin:
well i linked er up. hope my crazy commit names aren't too much for you xP

PR #2248, opened 2015-06-05
Icon for TigerVNC

nejsan:
Icon for [TigerVNC](https://github.com/TigerVNC/tigervnc).


Foggalong:
This would normally be against our guidelines because it doesn't use a symbol and instead makes use of the full canvas. Looking at the original icon though, and because I don't have any better ideas, I think this one will do. Swap out your black for `12d2d2dff`, it makes it a bit easier on the eyes :)

nejsan:
Right, I didn't see any guidelines and I wasn't sure what else to do. I saw a few other numix circle icons with elements clipping off the baseplate so I thought it would be ok if some of the stripes did. I also tried removing some of them but it makes the icon less understandable.

Foggalong:
Yeah, some icons are just made for clipping off the baseplate. Chrome and Firefox are both examples of it. Generally we ask first time contributors to stick to the guidelines, but I can see why you didn't in this case.

Anyways, it seems you mixed up the sed :P - `12d2d2dff` isn't a valid hexcode

nejsan:
Whoops! Also, where are these guidelines? I couldn't find any in the brief search I did.

Eleuin:
https://github.com/numixproject/numix-wiki/wiki/Guidelines (i got your back Josh)

Foggalong:
@nejsan No idea how you're doing this colouring but it's just not working :P

nejsan:
Sorry, I've just been manually editing the SVG. I'm not quite sure what 12d2d2dff is supposed to come out to in hex, so I assumed it was 2d2d2d.

Foggalong:
Yeah, `2d2d2d` is the correct colour code (the 1 at the front was a typo on my part) and Inkscape adds an extra two bits to control opacity, the `ff` part

nejsan:
Ah, ok, so it was an RGBA value with an extraneous first digit. I noticed that I forgot the `#` in the color codes, but that's fixed now. Sorry about all this!

Foggalong:
Not a problem! Thanks for the contribution :)

palob:
@nejsan Do you remember which icon you took the baseplate from? Thing is it was 47px X 47px big instead of 46x46 (already fixed it in https://github.com/numixproject/numix-icon-theme-circle/commit/7cb73cd4ce356dc7a6ca10ca4809eb51e7147cb3). Maybe there are some more icons with this issue?

nejsan:
No, I don't remember. I believe it was something with a red gradient and possibly had a symbol that was clipped off the baseplate. Possibly a steam icon.

PR #2256, opened 2015-06-08
Wolfenstein 3D icon

Eleuin:
should fix 2 issues for the two Wolfenstein 3D engines currently in the repos
wolf3d


Foggalong:
Switch the red for `900000ff`, it makes the contrast come across a bit less clashy :)

Eleuin:
no problem, i can commit this in a flash!

Eleuin:
slightly lighter than  900000ff, but it removes a lot of the colour clash

PR #2259, opened 2015-06-09
Icon for Monster RPG 2

Eleuin:
monster-rpg-2
monster-rpg-2-2


palob:
You need to move the baseplate down such that the "bevel shadow" is visible.

Eleuin:
yeah, i reapplied the symbol to another baseplate. i actually never noticed this, thanks btw

palob:
Oh no, you fell victim to the "gradient bug". The baseplate shows up correctly in Inkscape but is displayed red in file managers / image viewers.

To solve this, change the baseplate filling to a solid colour or to nothing and do the gradient from scratch. That's always the way to go when changing a gradient unfortunately.

Eleuin:
ok, that was weird... and definitely something i should remember for later. i've just never seen this before <.<

Eleuin:
@Foggalong any thoughts?

Foggalong:
I can't check SVGs very thoroughly on my tablet, which is where I do most of my labelling, commenting, etc. It's a very nice icon, though the symbol possibly needs moving down 1 or 2px to account fo

Eleuin:
yeah, its rather top heavy. i can move the symbol down a bit

Foggalong:
Excellent! :D

PR #2273, opened 2015-06-12
added icons and symbolic links

l0okme:
for: 9gag, drumstick, obs (open broadcaster software), slgui (sooperlooper) and xilinx ise


palob:
A few issues:
+ Generally the shadow needs to be 1px down and to the right relative to the symbol
+ On the drumstick icon convert the red stroke into a path, the remove the inner nodes. Don't have the drumhead circle reach the icon rim. This two measures will circumvent the jagged outlines. Furthermore the drumstick needs a shadow. 
+ Use white L249 instead of pure white
+ On the sooperlooper icon there are unneeded nodes and the logo colours could be brighter for a better contrast
+ Vertical and horizontal lines need to be aligned to the pixel grid as much as possible

Foggalong:
@l0okme A lot better! A few corrections still though:

* Drumstick is too far from the design guidelines to be merged and most likely needs completely redoing from scratch. If you want to just remove it from this merge and then have a separate one for it that's fine
* The 9gag root file needs to be `web-9gag.svg` with symlinks to `9gag.svg` and the Chrome icon

l0okme:
okay, I'll redesign the drumstick icon later :+1:

Foggalong:
@l0okme You missed out `web-9gag.svg` from the commit!

l0okme:
Ooops :D sorry

Foggalong:
Not a problem :D thanks for the contribution!

PR #2277, opened 2015-06-12
Icon for nautilus-scripts-manager

gregorydk:
Icon for nautilus-scripts-manager (http://www.pietrobattiston.it/nautilus-scripts-manager). It's a copy of `web-google-apps-script.svg` but with `file-manager.svg` colours.


palob:
It seems you ran some SVG cleaning here. At least it's not the unaltered Inkscape output.
If you do so, please use the latest script from [here](https://github.com/numixproject/numix-tools/blob/master/numix-tools/svgclean).

Your icon still has the "[gradient bug](https://github.com/numixproject/numix-icon-theme-circle/pull/2240#issuecomment-110940978)".

gregorydk:
I used `file-manager.svg` for the baseplate, I didn't check that icon for errors so if there's an error in that svg it got copied over. I use Scour (http://www.codedread.com/scour/) for svg cleaning since the svgclean script doesn't work on my distro.

palob:
Hmm, Scour doesn't seem to have an option to prevent the gradient bug.
The svgclean script needs svgcleaner(-cli) installed. Isn't that included in your distro?

PR #2282, opened 2015-06-15
added drumstick icon

l0okme:
The new drumstick icon is designed after the original one.


Foggalong:
Baseplate colour needs to be a bit lighter to make the contrast less harsh. Try a colour around `#8ad14fff`

palob:
@Foggalong colours are updated.

PR #2297, opened 2015-06-18
Fixed ktip baseplate and filename

palob:
The new ktip icon had the odd baseplate design from the redshift icon and there were 2 dots in the file name.


Foggalong:
That's weird, it was showing up right on mine. But thanks for the correction! The dots came from an Inkscape crash where it saved it to `ktip.$DATETIME.svg`

PR #2311, opened 2015-06-21
Added icon for KeepNote Note taking software

markubiak:
Used the background from Steam/DeadBeef, if I need to change it (as those are 2 popular apps) then let me know.  Tried to repliacte [KeepNote icon from their website](http://keepnote.org/images/keepnote-logo.png).


Foggalong:
Thanks for the contribution!

PR #2334, opened 2015-06-22
Added cool-retro-term.svg

dirtydancing:
Fixes https://github.com/numixproject/numix-icon-theme-circle/issues/2333.

cool-retro-term

### DIFF
--- a/Numix-Circle/48x48/apps/cool-retro-term.svg
+++ b/Numix-Circle/48x48/apps/cool-retro-term.svg
@@ -58,11 +58,11 @@
        y2="47"
        gradientUnits="userSpaceOnUse">
       <stop
-         style="stop-color:#7e7064;stop-opacity:1"
+         style="stop-color:#77695d;stop-opacity:1"
          id="stop7" />
       <stop
          offset="1"
-         style="stop-color:#73665b;stop-opacity:1"
+         style="stop-color:#6c5f54;stop-opacity:1"
          id="stop9" />
     </linearGradient>
   </defs>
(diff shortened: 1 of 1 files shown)


Foggalong:
While this isn't sticking to the guidelines strictly, I do really like how you've done the CRT effect so I'd be willing to merge. You need to fix the small problems with the shadow on the left hand side of the icon though :)

dirtydancing:
Yeah, this is not exactly sticking to the guidelines, but I really liked it this way, too, so I filed this PR :)

Which problems with the shadow on the left hand side are you referring to? I went for a drop shadow below the darker stripes, so the shadow "lines" are 1 px down and 1 px to the left.

As a side note: the darkest line is not crossing above the orange letters on purpose, because then the letters will become unreadable when scaling down. And there is no extra drop shadow for the orange letters, because there already is a "line drop shadow", and then the icon would then get overloaded.

This being said, could you point me to what shadow problems you are referring to?

palob:
You'd expect the shadow of the horizontal lines to be continuous and not to be following the baseplate's curvature.

dirtydancing:
Ah, ok, I did the following of the baseplate on purpose, but I see that continous shadows make sense (also goes for the very upper right then). Will correct that.

dirtydancing:
Corrected shadows:

corrected_shadows

Foggalong:
@dirtydancing Just to edge it a little closer to the guidelines, could we have a shadow for the CRT & █

dirtydancing:
@Foggalong Sure thing, will have a look into it, this might be a little while as I can't do it right now.

This might get somewhat tricky, as then probably the current shadow line cannot be above the letters anymore (otherwise there would be a shadow above and below). I did it the current way on purpose, to carry over the line effect over the letters without making them unreadable. Will have to see what I can come up with.

Eleuin:
I don't know if its just me or the "Symbol " is a tad bit too low. maybe move the CRT & █ up 1 px? That would also fix the scanline alignment on the symbol

palob:
Could be a 1px higher indeed. And maybe the letters a tad brighter & less saturated for more "glowing effect"?

Eleuin:
@palob my thoughts exactly

dirtydancing:
@Eleuin @palob Thanks for your feedback. I appreciate it and incorporated it in this enhanced icon version: Letters are 1px up and aligned to the scanline. The letter colors are slightly changed and now have an orange line to match the scanline and to enhance the "glowing effect". At the same time, the icon now scales even better in the smaller sizes. @Foggalong Drop shadow for the letters added.

enhanced-icon-version

palob:
:+1: However the shadow looks somewhat sparse. How would the shadow look as just duplicated letters in full black with 10% opacity? This way shadow would add up its opacity to all what's below.

dirtydancing:
Actually, the letter shadow was already duplicated letters with full black with 10% opacity, just sorted underneath the horizontal lines so as to create the impression of a drop shadow of the combination of the letters and the darkest line. Sorting this letter shadow right below the letters and above all the horizontal lines does make it somewhat more prominent and arguably more in line with the guidelines.

Good call! :+1: New version pushed.

enhanced-letter-shadow

Foggalong:
ooo, I like that

Eleuin:
@dirtydancing sexy, also the progress is nice to see :smiley:

dirtydancing:
Thank you all! I am glad to hear that you like it :)

dirtydancing:
Alas, one more thing I noticed: although the current design works very well on its own, there is a potential contrast problem with the background colors e.g. in the default GNOME dash on the left hand side in the activities overview. Such contrast problem does not exist e.g. with the lighter grey terminal icon. So I will have a look into a somewhat lighter background colouring.

dirtydancing:
OK, here we go: lighter background.

lighter-background

To give a visual impression of the difference, compare these two screenshots with Arch/GNOME, default Adwaita GTK theme/GS theme/wallpaper:

Before:

cool-retro-term_arch-gnome

Current:

cool-retro-term_lighter-background

Before, imo the icon was somewhat too dark, resulting in a sub-optimal contrast esp. in relation to the GNOME dash. With the lighter background current version, there is a better contrast, the circle shape is clearer, and the icon fits in better with the other icons overall, which results in more consistency.

palob:
I think now the letter's contrast is a bit weak again. But be that as it may.

dirtydancing:
@palob Agreed. Now with some more contrast and "glow" :)

more-contrast

Foggalong:
I'll be totally honest, the [high contrast one](https://github.com/numixproject/numix-icon-theme-circle/pull/2334#issuecomment-114860703) was pretty great.

palob:
Yeah, there are quite a few icons darker than the terminal one and the effect on the Gnome dash is acrually quite nice.

dirtydancing:
Yeah, there are darker icons around, the high contrast version from above is the one I like the best on its own as well, and also around the other icons that one makes for an engaging diversity. Pushed again :)

PR #2356, opened 2015-06-26
Added fontmatrix.svg

dirtydancing:
Fixes https://github.com/numixproject/numix-icon-theme-circle/issues/2355.

fontmatrix

### DIFF
--- a/Numix-Circle/48x48/apps/fontmatrix.svg
+++ b/Numix-Circle/48x48/apps/fontmatrix.svg
@@ -1,0 +1,414 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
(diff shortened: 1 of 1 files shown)


palob:
Cool idea but quite a heavy one. Is it any discernible at sizes below 48px?

dirtydancing:
Yeah, I checked the smaller sizes as well. Of course, the individual "a" letters of the "matrix" are not discernible anymore, but the shape/writing is (f/m, according to the original icon). It just turns into a sort of solid line once the size gets small enough :)

To illustrate, here is a screenshot with Arch/GNOME, larger size in the app overview, smaller size in the dash:

fontmatrix-screenshot

palob:
Maybe a lighter red to have the black part stand out a little more at small sizes?

dirtydancing:
Maybe. I checked a couple of different variants, and the problem with a lighter red was that then the contrast esp. to the right part of the "writing" was suffering. Anyways, I am surely happy to have another look at it :), might just be a little while.

palob:
The white is a bit harsh. It's L254 :)

dirtydancing:
Thanks @palob! Oops, I thought I had gone for Numix white #f9f9f9 :)

dirtydancing:
Lighter version: I checked a couple of variants with the starting point of the colours from above, and I was not really convinced by any. So this version is indeed "lighter" :) White is now Numix white #f9f9f9, and with the modifications suggested by @palob the icon size has now come down from 22 kb to 7 kb.

lighter-version

PR #2365, opened 2015-06-29
Lollypop enhancement

palob:
The current Lollypop icon hasn't got any shadow and is a little bit off from the original icon.
lollypop-old

Pushed:
lollypop

Alternative designs:

I hope it's okay, @ChuckDaniels87 :)


sergioad:
I love the pushed one :smile:

ChuckDaniels87:
Nice job, it looks better now. I like first alternative icon, a flat base to the candy (no shadows). I'd keep white circle shadow but remove the arrow one. :wink: :

dirtydancing:
Well, I for one have to say that I actually like the current icon better. It might just be personal preference.

It does deviate somewhat from the design guidelines and from the original icon, but why not?

For some reason, the new version with the shadows esp. for the center circle and for the outside pattern reminds me of `help.svg`. Then again, I do not have a better proposal, so I will be happy either way :)

help

PR #2384, opened 2015-07-07
Removes weather-clear.svg from circle

wa4557:
Indicator-weather uses this icon to display the weather condictions in the panel.If this icon (which is clearly a status icon is present in circle) the circular icon is used instead of the status icon -> Remove it, and blame the program that uses an status icon as app icon

### DIFF
--- a/Numix-Circle/48x48/apps/weather-clear.svg
+++ b/Numix-Circle/48x48/apps/weather-clear.svg
@@ -1,1 +1,0 @@
-gnome-weather.svg


Foggalong:
Interesting history point: this icon is _old_. It was added to the theme some time between the last commit to @cldx original [Numix](https://github.com/cldx/numix) in Aug 2013 and the [initial commit](https://github.com/numixproject/numix-icon-theme/blob/49c80451bc37ff7ac5cc61db88aa8bdf482ec860/Numix/apps/48/weather-clear.svg) to the new Numix Project icon theme in Oct 2013. At that point Base was combined with what is now Shine in a single icon theme package rather than the modular approach we now use.

Foggalong:
Because it's a symlink it's also one of the few icons that has never been changed, only moved. It went from Numix to Shine to Bevel to Circle but the symlink was still the same symlink that was made nearly 2 years ago. There aren't any icons older than it because the original Numix theme [didn't contain any app symlinks](https://github.com/cldx/numix/tree/master/apps/48) until the 2 months transition period when it was switched to Numix Project.

PR #2401, opened 2015-07-10
Replaced strange objects in palemoon.svg

palob:
Fixed ``palemoon.svg`` which AFAIK was the last one containing those strange objects

Looks slightly different though, before/after view:
palemoon-old
palemoon


Foggalong:
I can't remember my process for making Palemoon but I don't remember it. Ring especially rigorous. Thanks for the fix!

PR #2407, opened 2015-07-16
Clipped path to intersection conversion

### LINKED ISSUE #2016: KDE breakage

palob:
Fixes #2016

Converted all clipped paths into intersections. Viewing in Gwenview and Nomacs there is no broken icon remaining as far as I can see.
YAY! :+1: :+1: :+1: Numix Circle is 100% ready for KDE / Qt now!

Side effects:
+ file size reduction
+ fixed glitches
+ reduced "librsvg outlines"


Foggalong:
Awesome! \O/

PR #2449, opened 2015-08-08
Replaced scudcloud.svg

markubiak:
I was unaware that there was already a Numix icon for ScudCloud, and made my own.  Since it is already complete, I thought I would create a pull request and see if the Numix project leads prefered mine or the existing icon.


palob:
A few small issues: You did a path combination instead of a path union and the horizontal cloud baseline isn't aligned to the grid. The baseplate colour looks a bit harsh.

That said I think I prefer the existing design but then again I don't use the program.

markubiak:
That's fine, I just thought I'd ask

palob:
I'm not to decide :)
Let's wait for other people drop by.

Foggalong:
We're trying where possible to move away from all white symbols so probably not. But thanks for the contribution!

PR #2480, opened 2015-08-14
Icons/Symlinks for Metro series

Eleuin:
Original:
metro
Pushed:
metro_numix

### DIFF
--- a/Numix-Circle/48x48/apps/steam_icon_286690.svg
+++ b/Numix-Circle/48x48/apps/steam_icon_286690.svg
@@ -1,0 +1,1 @@
+steam_icon_43110.svg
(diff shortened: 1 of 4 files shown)


palob:
Symbol needs to be 1px up :)

Eleuin:
well, i just used the Metro: Last Light icon, adding the symlink for the redux version and then recolouring for Metro 2033 (why didn't the original ever get one? why just the sequel?), but if we want to update both, then ill have a look

palob:
Probably because nobody filed an issue for the original one.

Eleuin:
either way, it does seem that the symbols weren't aligned properly, ive gone ahead and fixed that and did some cleaning up

PR #2497, opened 2015-08-23
Brackets alignment fix

palob:
The brackets weren't equally wide.

Before:
brackets-old

After:
brackets

Or better this way or with a shadow?
brackets0
brackets-shadow
brackets0-shadow


Foggalong:
Pushed is fine :+1:

Issue #450, opened 2014-03-22
Icon for GNOME Weather


Milozzy:
A numix-circle-icon for Weather is missing.
The icon line says "org.gnome.Weather.Application" and the original icon is attached
![schermata da 2014-03-22 14 43 38](https://f.cloud.github.com/assets/864646/2490823/03284742-b1c8-11e3-9a75-61614410c387.png)

Project's website: https://wiki.gnome.org/Apps/Weather

PR #2518, opened 2015-09-01
Removed non-app icons. Fixes #2412

### LINKED ISSUE #2412: Are info.svg and help.svg app icons?
Or does any application use it for its launcher icon?
You've probably seen something like this where the Circle icons ``help.svg`` and ``info.svg`` show up where they shouldn't.

help
info

We haven't got action icons named ``help`` and ``info`` yet.

Just to show it's really ``help.svg`` and ``info.svg`` this is how it looks after removing ``info.svg`` and renaming ``2048.svg`` to ``help.svg``:
help2
info2

@bil-elmoussaoui Was this what caused the issue you reported [here](https://github.com/numixproject/numix-icon-theme/issues/190#issuecomment-89869828)?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24522154-are-info-svg-and-help-svg-app-icons?utm_campaign=plugin&utm_content=tracker%2F514195&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F514195&utm_medium=issues&utm_source=github).
</bountysource-plugin>

palob:
Replying "No" to #2412.

I removed
``appointment``
``help``
``help-about``
``help-contents``
``help-info``
``info``
``stock_about``
``stock_help``
``stock_insert-note``
``stock_task``
``stock_unknown``
``stock_view-details``

Renamed ``snap`` to ``snap-icon``
Fixed all symlinks.

I didn't touch
``browser-help``
``stock_calendar``
``stock_help-agent``
``stock_notes``
``stock_tasks``

So these are to be watched further on.


palob:
OK, I think ``stock_addressbook``, ``stock_calendar`` and ``stock_notes`` need to go as well.
Evolution is a good app to see them show up as action icons.
Likewise ``evolution-taks`` might not be an app icon.

All but ``stock_notes`` need to be added to the base theme first though.

wa4557:
But all of these icons do exist already (addressbook, calendar notes) symlinks are sufficient I guess

palob:
Ah OK, it's not that easy with Evolution. Using Numix base only shows that it uses mimetype icons. 
Anyway, ``go-today`` most probably isn't needed in Circle.

I'll add the symlinks later.

palob:
@wa4557 ``stock_people`` for ``stock_addressbook`` and ``view-calender-day`` without the day highlighted for ``stock_calendar``?

wa4557:
There's a `stock_new_addressbook` Without the plus it would be a good `stock_addressbook`. The calendar icon I would do as you suggested

palob:
But ``stock_new-addressbook`` is just one person and looks like ``gtk-orientation-portrait``

palob:
@wa4557 I've done them but I'm yet to figure out whether they're really action icons.

wa4557:
Great. Well I guess all those `stock_` icons are actually `action` icons

palob:
But I haven't found them on any other theme. So maybe application-specific.

wa4557:
I thought they turn up in `evolution`?

palob:
They're actually mimetype (those strange ones ``x-...`` or ``gnome-...``) or real app icons. Nasty but it's with Adwaita in mind where the design for action and app icons doesn't differ distinctively.

wa4557:
So you suggest removing them altogether? I'd say that they don't do any harm. If you already have them I'd add them

palob:
Additionally there's a bunch of ``evolution-....`` icons in circle which are actually action icons.

wa4557:
I really don't know why they were added in the first place...

palob:
Maybe in the very beginning Base and Circle weren't separated?

wa4557:
I guess they weren't but still, why makng those icons circular? Doesn't make sense these icons are still action icons

palob:
By the way, Evolution uses ``stock_check-filled`` from https://github.com/numixproject/numix-icon-theme/issues/450.

wa4557:
are those `*mail*` icons from [#450] (https://github.com/numixproject/numix-icon-theme/issues/450) used as well?

palob:
I don't think so.

palob:
@wa4557 I've noticed that the ``evolution-...`` icons are present as app icons in the ``gnome`` icon set (which is a mistake but action icons and app icons are designed in the same way there).

So the non-app icons in Circle most probably were created in analogy to existing themes.

wa4557:
And how do they show up in evolution? Are circle icons appropriate there, or would it be better to have base icons for them?

palob:
I'll test that later. Using the latest masters it's consistent with not too small Circle icons in the sidebar and base icons in the toolbar.

wa4557:
I just tested it myself: I guess it looks OK in the sidebar, however the same icons also show up in the preferences where base and circle icons are mixed (`evolution-calendar` for example)

palob:
I'll try to make it use action icons. The mimetype icons when using Numix base only are strange however.

wa4557:
Indeed. The `drafts` entry in the sidebar is also using a circle icon...

palob:
Argh, the drafts icon might be a toxic one.

wa4557:
Kind of: The icon used is `accessories-text-editor.svg` which is arguably also not an app icon

palob:
That's definitely an app icon (e.g. used by gedit). If it was ``applications-...`` it would sound like a categoty icon.

wa4557:
Yeah I was wrong, gedit for example is using this icon :( That's an app icon -> evolutions fault

palob:
Btw, it's interesting to explore ``/usr/share/evolution/icons`` in this regard.

wa4557:
Ah interesting that's where all those png icons are coming from

Foggalong:
In the beginning. the base and apps were indeed all bundled in [one theme](https://github.com/cldx/numix). Then the apps were what is now known as Numix Shine, and general design was mostly based around single coloured shapes. The only parts of the theme which has been unchanged from that design are the category icons.

The unified development ended so that the same base icons could easily be used for Shine, uTouch, and the newly created Circle.

PR #2539, opened 2015-09-08
Icon for Ring

palob:
FLOSS video communication software, SFLPhone successor.
http://ring.cx/

**Original icon:**
ring-ori

**Pushed icon:**
ring


Foggalong:
I love it! :D

PR #2580, opened 2015-10-01
All the deleted commits from master

wa4557:
Fortunately I still had the missing commits in my master branch. So here are the deleted commits


Foggalong:
Legend :+1:

PR #2597, opened 2015-10-11
Icon for Otter Browser

palob:
web browser
http://otter-browser.org/

It's very close to the original, maybe a tad too detailed?

otter48
otter144


Foggalong:
It's close to the original, but isn't to departed from some of the other animal styles we've used :)

PR #2606, opened 2015-10-15
Adds hidpi support for numix-circle

### LINKED ISSUE #2377: Implement Better HiDPI Support
The brilliant folks at elementary recently went to a summit to talk to the GNOME folks. They talked about various things including increased collaboration (yay!) but the thing important for Numix is that they discussed how to implement better HiDPI support in their icon themes. This is the [relevant commit](https://code.launchpad.net/~elementary-design/elementaryicons/hidpi-test/+merge/263317) on Launchpad for the elementary icon theme and we basically need to dig through this and find what we need to add to our themes. 

I'd really like to finish #190 first since adding support for both involves significant changes to the `index.theme` file. The equivalent issue for Base is [found here](https://github.com/numixproject/numix-icon-theme/issues/610).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/23618143-implement-better-hidpi-support?utm_campaign=plugin&utm_content=tracker%2F514195&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F514195&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #2603: Does a HIDPI implementation make sense here?
1. First off: I guess we need to add a proper HIDPI implementation here in circle as well at least for the `panel` folders, but:
2. I'm unsure if the `48@2x` approach is necessary for the apps folders as well, because there is only one size after all. What do you guys think?
3. I guess we should at least make the naming (`48x48` -> `48`) consistent
4. Is the `scalable` folder really necessary, couldn't this be easier made with a `MinSize` and `MaxSize` for the `48` sized folder? (Doesn't matter that much actually)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27416525-does-a-hidpi-implementation-make-sense-here?utm_campaign=plugin&utm_content=tracker%2F514195&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F514195&utm_medium=issues&utm_source=github).
</bountysource-plugin>

wa4557:
HIDPI support for numix-circle. I removed the `scalable` folder and replaced it with a Minsize and maxsize for `48/apps` (consistent with numix-base). Fixes #2377 and fixes #2603. With this numix has full HIDPI support :)

On another note: A working ppa would be really cool now...


Foggalong:
Before I merge this I need to bring Square up to date too so that it also uses

bil-elmoussaoui:
Do you think that's possible to have a bash script that will change circle icons to square ones?

Foggalong:
@bil-elmoussaoui We had one previously which is what spurred the creation of many other Numix themes (Hexagon and Fold being the only formally released of those). It broke beyond repair when librsvg was updated though. The same thing had happened before - basically the libraries required to make it all work change too frequently to setup a stable system.

wa4557:
Is the script available somewhere? I'd be interested to see how it is done

bil-elmoussaoui:
It's might be possible to update the script and fix all the errors, and keep updating the script whenever librsvg gets a new update?
I would like to take a look on it too! as i really want to use Numix Square but a lot of icons are missing, and adding them manually will take a lot of time!

Foggalong:
It wasn't my preference but it was closed-source. I'll speak to the guys about opening it up. Just to give you a quick insight into a few of many of the things that needed fixing for it to be usable (from the README):

* No longer works with the librsvg tools in 14.04+
* Needs to use recursive intersecting instead of clipping (quite a few DEs don't support clipping)
* Native gradient support, rather than a filter layer or hacky fix

Foggalong:
[Here it is :)](https://github.com/numixproject/script-o-mighty.git)

palob:
Anyway in order to fix the Qt bugs I transformed all clipped paths into intersections for all respective Circle icons. As a result the background symbol paths are now to small for the Square icons.

Foggalong:
That's fine :) when I was constructing Square in the first place I had to deal with that problem so it's not too bad

palob:
If the script started to break with Ubuntu 14.04 [it's probably](https://launchpad.net/ubuntu/+source/librsvg) librsvg 2.40 which introduced the API break.
If the change can't be worked against how about running the script after booting into a 12.05 ISO?

bil-elmoussaoui:
@palob i can try this if you want to ? :)

wa4557:
@Foggalong You should merge all the other PR first, since changing them would be more work than changing this here. Since the changes are really easy to make, I can open a new PR as soon as you have finished square :) I will close this one

palob:
@bil-elmoussaoui Feel free to go ahead. I haven't looked into what the script collection does so far.
Is there one for creating Square icons from Circle icons?

Foggalong:
The script makes both the Circle and Square (and unreleased Hexagon) icons out of a collection of symbol only icon

PR #2623, opened 2015-10-21
New Turpial Icons and Evolution more actually

Nebrit:
Two icons change proposed. Turpial the updating of the official logo and colred; much more attractive and icon of evolution that resembled the Thunderbird, it has been redesigned with a new concept that fits all platforms.


palob:
Thank you that you want to contribute however your icons really aren't in line with the [guidelines](https://github.com/numixproject/numix-wiki/wiki/Guidelines). 

Apart from what's pointed out there don't put anything above the baseplate like you did in your Turpial draft. The colours need to be applied to the baseplate itself.

+ Don't use fully saturated colours, the yellow one is too garish.
+ Use white L249 at the brightest instead of full-blast white.
+ Use something like dark grey L40 at the darkest instead of full-blast black.
+ The icon should be somewhat recognisable at smaller sizes (``<48px``). The GNOME foot however isn't even crisp at 48px. It isn't ideally positioned / aligned to the pixel grid.

Just open a few existing icons in Inkscape and examine how they're composed.
If you've got questions on how to do this and that in Inkscape you can ask here.

What's that orange thingy in your Evolution icon for? Just recolour what's below.

Nebrit:
Thank you.

I have some doubts, are the following.

It ordered the mistakes I've made as each icon I uploaded understand that looks like this:

Turpial:

- Don't use fully saturated colours, the yellow one is too garish.
- Use white L249 at the brightest instead of full-blast white.
- Use something like dark grey L40 at the darkest instead of full-blast black.

Evolution

- The icon should be somewhat recognisable at smaller sizes (<48px). The GNOME foot however isn't even crisp at 48px. It isn't ideally positioned / aligned to the pixel grid.

In addition to responding to the fact recolor and add the symbol of Gnome it is because they are so similar current thunderbird and evolution icons that are almost indistinguishable. Hence the change of colors and add the symbol of Gnome. However seeing the minimum size of 48px I will change the icon by the clock of evolution positioning on the original site.

Gran duda !!
Excepto estos cuatro puntos, la composición restante es correcta y te gusta ??

I will upload all changed according to your specifications.

Thank you, I'm starting, excuse my doubts

Nebrit:
I uploaded another version of the two icons following the guidelines and changing colors so they are not as damaging to the eye.

The turpial think of right now.

The Evolution've redesigned according to what I see is the line design Numix.

If fol being wrong please tell me what I should rectify of the icon and perform.

Thanks.

palob:
Sorry but there are still a lot of clashes with the guidelines.

Foggalong:
Yeah, not a fan of either of these redesigns. Sorry!

PR #2624, opened 2015-10-22
Icon for MAME

gregorydk:
Fixes https://github.com/numixproject/numix-icon-theme-circle/issues/2579


Foggalong:
Good job! Could you add a background gradient from `d6ecf6ff` to `e6f4faff`, just adds some variety

PR #2657, opened 2015-11-12
Add icon for Duke Nukem 3D: Megaton Edition (Steam)

MrTimscampi:
Steam ID: 225140

Also available on GOG (Both Megaton Edition and Classic Edition), so, before merging, someone should find the names of those icons and symlink them with this one.

Icon:
steam_icon_225140


bil-elmoussaoui:
I think that you `master` branch contains all the other icons :s I don't know if you can separate them now! but you can remove your pull requests, remove your fork and re-fork again, and create a new branch for each icon!  
Nice icon! keep it up!

MrTimscampi:
Yeah, I didn't think it would pull all the other commits ^^

I cleaned up my branch by removing all my previous commits from it and force-pushed to the branch, seems to have worked.

palob:
Small suggestions for this one:

Can you reduce the number of nodes and turn this
atom1
into this?
atom2

The black and yellow elements look a bit harsh. Can you try to turn up lightness and reduce saturation (yellow only)?

Foggalong:
screenshot_2015-11-15_15-31-46

This horizontal line really needs to be aligned to grid. Also, our size for circular symbols is 30px to average things out appropriately :)

Foggalong:
What going on with the freaky node positioning here?
screenshot_2015-11-21_14-20-29

palob:
To get the sign properly aligned just do it from scratch this way:
+ Put 3 circles (26px X 26px) on the yellow circle positioned at (11,11)
+ Use the circle tool to turn them into segments: one from 60° to 120°, one from 180° to 240°, one from 300° to 360°/0°
+ Put a 8px X 8px circle in the icon's center and merge it with the circle segments
+ Put a 4px X 4px circle in the icon's center and do a path difference

atom
atom2

Foggalong:
@palob what do we use for SVG cleaning these days?

wa4557:
We're using something special for SVG cleaning oO You should have told me that... I' always just using inkscape --vacuum-defs

palob:
Inkscape's vacuum-defs (GUI or CLI does the same afaik) and the ``svgcleaner-cli``-based [svgclean script](https://github.com/numixproject/numix-tools/blob/master/numix-tools/svgclean). 
@gregorydk uses [Scour](https://launchpad.net/scour). Seems to work as well if you watch out to use the right options.

Is your question related to the above screenshots? I didn't use any sort of cleaning method there.

Foggalong:
Ah okay. I can't use the latter because it doesn't support my Ubuntu version so I'll just make do with vacuum-defs. And yeah, GUI & CLI run the same command :)

gregorydk:
Inkscape uses Scour if you choose to save it as Optimized SVG.

palob:
Thanks, that was unbeknownst to me. And you can check the options from the GUI.
Do they bundle Scour? The package isn't installed on my system.

gregorydk:
@palob Inkscape includes it as an extension. You can also install it as a python script, it's called `python-scour`. Should be in the Ubuntu repos.

PR #2660, opened 2015-11-12
Added icon for Shadow Warrior Classic Redux (Steam)

MrTimscampi:
Steam ID: 225160

Also available on GOG as Shadow Warrior Classic Complete. That icon should be symlinked to this one, but I don't have the name for it.

Icon: 
steam_icon_225160


bil-elmoussaoui:
You can find the steam id by looking for the game using your default browser, for example : http://store.steampowered.com/app/238070/?l=french , the steam id is 238070 ;)

MrTimscampi:
Didn't know Classic was also on Steam, will make the symlink now :)

My comment was about the GOG version, though. They have a different naming scheme for icons.
Their icons seem to be hardcoded, though (I only have Pillars of Eternity to test that), so it might be okay to only have hardcode-fixer rename the icon to steam's icon.

palob:
Some advice to counteract the infamous librsvg outline bug:
Currently your icon looks this way:
outline
There's this jagged outline around the black part. This happens when two objects are placed one above the other and share the outline.

Reducing the invisible part of the white object such that it doesn't touch the black part's outline fixes the outline.
outline1b
no outline1

Result:
nooutline

MrTimscampi:
Will fix it and make a new commit, thanks :)

Foggalong:
While I get what you're going for, I think it stands out a bit too much to be merged as is. Not sure how you'd made the symbolism work with our shadow guidelines. Might a do over :(

palob:
Shadow for the lower hole and the blood outline? Shadow for the upper black shape on the white shape like this (modified ``kreversi.svg``):
kreversi

Foggalong:
I stand corrected, that actually looks really awesome.

MrTimscampi:
Tried adding a shadow to the existing design:
shadow

Foggalong:
Assuming @MrTimscampi has vamoosed, would you be okay to pick this up @palob? I really like what you did with making the whole Yin Yang a symbol and how that worked shadow wise :)

palob:
I could do this as proper yin-yang symbol however [here](https://steamdb.info/app/225160/) it looks as awary as @MrTimscampi did it. Well, maybe it was their artist not doing it properly.

Foggalong:
It's possible yeah. Gonna close this pull anyway since @MrTimscampi vanished and it needs a bit of redesign before it could be merged anyways.

PR #2661, opened 2015-11-13
Icon for NVIDIA Visual Profiler

MrTimscampi:
Fixes numixproject/numix-icon-theme-circle#1901


palob:
There are 2 small issues: 
+ The label isn't properly aligned to the pixel grid (can be toggled by pressing the ``#`` key in Inkscape) which will make it look blurry.

+ The upper and lower line of the graph aren't parallel where it hits the label, looks a bit askew.

PR #2714, opened 2015-11-28
Tkpacman symlinks, yumex symlink enhancement

palob:
GUI for the pacman package manager
http://sourceforge.net/projects/tkpacman/

**Original icon:**
tkpacman.desktop
Icon=/usr/share/tkpacman/icons/tkpacman-icon.png
tkpacman-icon

Furthermore I've edited tje yumex symlinks such that the yumex icon is more close to the original icon and looks distinct from the icon used by gnome-packagekit.


Foggalong:
Like your thinking :+1:

PR #2754, opened 2015-12-30
Add goa-account icons

forivall:
I created the live icon from the numix "live" icon and the background of the microsoft icon. It probably needs a shadow added. I created it using inkscape & `svgo --pretty --disable=collapseGroups,convertPathData live.svg`

Otherwise, I just set up the appropriate symlinks.

### DIFF
--- a/Numix-Circle/48x48/apps/goa-account.svg
+++ b/Numix-Circle/48x48/apps/goa-account.svg
@@ -1,1 +1,0 @@
-goa-panel.svg


palob:
Hi, where are those icons supposed to show up? They're part of the [https://github.com/numixproject/numix-icon-theme](Numix base theme).

forivall:
For some reason (it would be nice if someone could test on distros other than Arch), the numix base icons don't show up. These appear in the "Online Accounts" preference pane of gnome-control-center.

wa4557:
See https://github.com/numixproject/numix-icon-theme#847

bil-elmoussaoui:
@Foggalong This should not be merged, it's an issue with Gtk. The icons do exists in Numix but again, Gtk uses `goa-account` that exists in Numix Circle :/

forivall:
Numix base OR with goa-account.svg deleted (note that goa-account is just a symlink to goa-panel.svg)  screenshot from 2015-12-31 13-33-02
This pull request: screenshot from 2015-12-31 13-34-40
Stock / Adwaita / hicolor: screenshot from 2015-12-31 13-35-16
Before this pull request: screenshot from 2015-12-31 13-39-00

palob:
OK, so even if it would look better with non-Circle icons your PR improves on status quo.

bil-elmoussaoui:
I don't think that we should make goa-accounts* circle..

forivall:
Yeah, so if we really don't want circle'd goa-accounts* icons, I'll change this pull request to just remove goa-account.svg, and then submit a PR to numix base to add lastfm, exchange and fix the microsoft account icon (goa looks for goa-accounts-msn, not goa-accounts-live)

bil-elmoussaoui:
`go-account` is used somewhere i guess; try removing it and use Numix Circle and see if you will have any missing icons?

forivall:
goa-account.svg is only used within the account selection screen as shown above. The icon to get to the goa-accounts preferences is goa-panel.svg. The first image in the above comment is what it looks like with goa-account.svg removed. The hicolor version of goa-account.png is just a transparent png.

bil-elmoussaoui:
@dirtydancing Can you test this please ? If we can remove goa-accounts from circle this will be really cool. We won't need to create any symlink to fix those icons.

forivall:
For reference:
http://standards.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#icon_lookup
https://github.com/GNOME/glib/blob/de04fd13048dd208162573187e4c0d9e7d3428d3/gio/gthemedicon.c#L344
https://github.com/GNOME/glib/blob/de04fd13048dd208162573187e4c0d9e7d3428d3/gio/gthemedicon.c#L150
https://github.com/GNOME/gtk/blob/861ac8e373681cacfe577b1d6c3c17934a33e1aa/gtk/gtkicontheme.c#L1883

Gtk makes the lookup for theme-specific fallbacks before moving on to checking parent themes. So if we wanted to fallback to the numix icons while still overriding a default fallback, you would need to symlink all of the more specific icons from numix into the numix-circle theme folder.

bil-elmoussaoui:
Screenshot after removing `goa-account` in Gnome 3.18.3 using Fedora 23
screenshot from 2015-12-31 23-25-06
And everything looks fine
screenshot from 2015-12-31 23-25-22

bil-elmoussaoui:
I confirm that this fixes https://github.com/numixproject/numix-icon-theme/issues/847

wa4557:
On my machine numix is the only theme which supports this icon. So I suppose it is OK to remove it (i.e. everything fine in Ubuntu whenremoving it)

dirtydancing:
Alright, this looks like an inheritance line GTK lookup situation similar to the classic https://github.com/numixproject/numix-icon-theme/issues/190.

I agree that removing `goa-account.svg` from Numix-Circle is the way to go, and I can confirm that removing `goa-account.svg` (symlink with the target `goa-panel.svg`) from Numix-Circle does not cause breakage with Arch/GNOME, as the icon used in system settings is `goa-panel.svg`.

PR #2757, opened 2015-12-31
Maintenance

palob:
Fixed misalignments, outlines, baseplates, missing shadows and other small issues. Merged shadow paths.


Foggalong:
Just at a first quick glance through, it seems to have done something weird to the hole colour of the purple calendar.

palob:
Oops, botched something and didn't fix what I actually wanted to. 
Fixed now.

PR #2782, opened 2016-01-09
Fixes #2780

### LINKED ISSUE #2780: Genymotion Player
Genymotion uses a different icon for it's emulator player
The icon name : the desktop file is not created. We can use `genymotion-player` 
The default icon : the same as #1205 with inverted colors 
genymotion

palob:
Fixes #2780 
genymotion-player


bil-elmoussaoui:
That was fast :p Thanks!

PR #2792, opened 2016-01-15
Add discord icon

forivall:
It doesn't have an official linux app yet, and the unofficial app (at least the .desktop file in the current AUR package) [uses a hardcoded path for the icon](https://aur.archlinux.org/packages/discord/#comment-528810).

But hey, why not add it.


Foggalong:
Looking good! Just as a small correction, we use `#f9f9f9ff` instead of `#ffffffff` for white :)

Foggalong:
Also in the gradient, only change the L value - leave S constant

PR #2808, opened 2016-01-25
Icon for veracrypt. Fixes #2803

### LINKED ISSUE #2803: Icon for VeraCrypt
VeraCrypt is based on TrueCrypt(which has no longer support).
captura de pantalla de 2016-01-24 17-42-51

zsteinmetz:
I used contours to avoid the tiny gaps between triangles.

### DIFF
--- a/Numix-Circle/48x48/apps/veracrypt.svg
+++ b/Numix-Circle/48x48/apps/veracrypt.svg
@@ -1,155 +1,161 @@
 <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48"> 
   <defs
-     id="defs4" />
+     id="defs4">
+    <linearGradient
+       id="linearGradient4382">
+      <stop
+         id="stop4384"
+         offset="0"
+         style="stop-color:#164b6e;stop-opacity:1;" />
+      <stop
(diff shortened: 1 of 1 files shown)


palob:
Like the colours but using strokes is discouraged, it poses a lot of issues and contributes to a glitchy look. A better solution for the gap issue would be overlapping tetragons.

That aside the symbol is a bit big (max regular size is 26x26 or 30x30 for circular ones) and horizontal/vertical lines need to be aligned to the pixel grid (toggled with ``#``).

zsteinmetz:
I understand. Hope its better now.

palob:
Better now, but there are still horizontal edges non line with the grid.
work

It might not be easy to fix this with the current design but then again it might be too detailed currently anyway. The symbol isn't easily recognisable at smaller sizes. Can you think of a slightly simpler design involving less triangles?

The gradient is supposed to be aligned this way:
gradient

palob:
:+1: Everything nicley aligned now. Might be still a bit busy, @Foggalong?

As for the width I think the symbol works still when squeezed to 26 px even if the triangles aren't equilaterial anymore.

The guidelines definitely could do with a makeover / some pics.

PR #2847, opened 2016-02-08
Distinct icon for Sky Chart

palob:
I've had this resting on my disk for a while:
Diverging from the guidelines but recognisable and small sizes and true to the original icon:
skychart

Alt with shadows:
skychart-shadow

Maybe a bigger meteor?


ChuckDaniels87:
It diverges from guidelines but it is really nice icon. Could you try bigger meteor?

Foggalong:
While I think it's a nice icon, I'm not going to merge it because it's too far from the guidelines.

palob:
That's actually a good idea, will try it.

palob:
So something like this?
skychart2

Foggalong:
Yeah, that's much better :)

palob:
Pushed it :+1:

PR #2850, opened 2016-02-08
Cleanup+opacity fixes for grsync and timeshift

palob:


bil-elmoussaoui:
I think that the extension of grsync need to be removed?

palob:
If it was just made to fix a hardcoded ``.png`` path (which afaik doesn't work this way), yes.

There's already a symlink for the pure ``.svg`` icon name though.

PR #2854, opened 2016-02-12
Create cisco-anyconnect.svg

r0ckarong:
Created icon from scratch with some inspiration.

cisco-anyconnect

This would fix https://github.com/numixproject/numix-icon-theme-circle/issues/2820


mnapp:
How does the actual integration work? The svg is not named like the original icon and it's obviously not a png icon file. Do you guys render this out with the correct name according to the icon request? If i can help with this at all let me know.

palob:
You can copy the ``.desktop`` file to ``~/.local/share/applications`` and edit the icon line to ``Icon=cisco-annyconnect``.

Alternatively, you can use the [hardcode-fixer](https://github.com/Foggalong/hardcode-fixer) script. Please report the hardcoded icon path against this repo or open a PR fixing it.

mnapp:
@palob I've seen some of your icons here. Do you have any pointers why mine looks too small no matter what size I export it to? I tried everything from 48 to 128 but it always looks a bit "off" in the Unity launcher. What is so different to the other Numix icons? Sorry I'm totally new at vector graphics.

palob:
@mnapp For some reason you question seems to have vanished, new contributors are always welcome.

It looks like you have done the baseplate from scratch. Just start from an existing icon, delete the symbol and recolour the baseplate.

The reason for your impression of wrong size/misalignment arises from your symbol not being aligned to the pixel grid (toggled with the ``#`` key in Inkscape).
pixelgrid
If a vertical/lateral border of two differently coloured regions isn't in line with the pixel grid a line with interpolated colour will be rendered in between. The result is a blurry shape.

mnapp:
Sorry yeah, I removed that comment because I figure out how to set the SVG scalable in the .desktop file and then it looked right. I've now taken your advice and recreated the icon from an existing backplate and properly aligned it with the grid. I also ran svgcleaner to make it more compact. Thanks for the help.

palob:
Probably due to svgcleaner arguments you ran there's strange stuff in your SVG's source code and the symbol object exceeds the baseplate.  It looks like you've done the shadow using a filter. Just duplicate the symbol shape and make it black with 10% opacity.

That's the only reasons I can think of why it would appear bigger.

When you export from Inkscape use "page". Anyway, the Inkscape raster export is a PITA imo.
Alternatively you can for instance open the SVG in GIMP, render at 48x48 and export to PNG.
Then again you shouldn't need to go the PNG way at all.

palob:
I've had another look at the icon.

We've got [those design guidelines](https://github.com/numixproject/numix-wiki/wiki/Guidelines). Yes, absolutely hidden and not thorough.

Right now your symbol is a bit too wide. Furthermore it isn't exactly symmetrical and some lines are still misaligned. Can you recreate the symbol using a combination basic shapes like (rounded) rectangles and circles? You'll need to use path operations like ``Union``, ``Difference``, ``Intersection``.....

mnapp:
I ran svgcleaner-gui with default settings so no idea how to correct this. As I mentioned before I have no real experience with vector graphics before this. I know how to use paths somewhat in GIMP and others but I need to spend more time to actually work with it in Inkscape.

I will give remaking the symbol a shot in the next couple of days.

palob:
For cleaning up SVG icons you can use [this script](https://github.com/numixproject/numix-tools/blob/master/numix-tools/svgclean). It runs ``svgcleaner-cli`` without the harmful options.

PR #2868, opened 2016-02-25
Icon for LTSpice

Jacajack:
I recently created icon for LTSpice, a electronics simulation software.
It doesn't run on Linux, but You can easily use it in WINE, so I thought You may want to consider adding its icon to the set.

This is my Icon:
selection_527

*(Again, sorry for that brown backround)*

I tried to follow guideline, but if something's wrong, let me know, I'm willing to fix it. ;)

### DIFF
--- a/Numix-Circle/48x48/apps/ltspice.svg
+++ b/Numix-Circle/48x48/apps/ltspice.svg
@@ -41,8 +41,8 @@
      id="namedview65"
      showgrid="false"
      inkscape:zoom="11.313709"
-     inkscape:cx="15.501814"
-     inkscape:cy="30.530397"
+     inkscape:cx="17.007853"
+     inkscape:cy="31.818551"
      inkscape:window-x="65"
      inkscape:window-y="24"
      inkscape:window-maximized="1"
@@ -57,11 +57,11 @@
     <linearGradient
        id="linearGradient3788">
       <stop
-         style="stop-color:#dbd5d1;stop-opacity:1;"
+         style="stop-color:#dfcdcd;stop-opacity:1;"
          offset="0"
          id="stop3790" />
       <stop
-         style="stop-color:#e3dddd;stop-opacity:1;"
+         style="stop-color:#e6dada;stop-opacity:1;"
          offset="1"
          id="stop3792" />
     </linearGradient>
--- a/Numix-Circle/48x48/apps/ltspice.svg
+++ b/Numix-Circle/48x48/apps/ltspice.svg
@@ -5,6 +5,7 @@
    xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
    xmlns:svg="http://www.w3.org/2000/svg"
    xmlns="http://www.w3.org/2000/svg"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
    xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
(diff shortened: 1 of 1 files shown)


palob:
While this one might be a case for an exception to the rule the symbol is to big currently.
I think a size 30x14 works well. If centered it'll also be aligned to the pixel grid then.

Can you use a slightly lighter and less saturated colour on the symbol and a light reddish vertical gradient for the baseplate (10L lighter on top)?

Jacajack:
Is that better now?
selection_538

palob:
Yep :+1: Maybe a bit more red for the baseplate?

Jacajack:
I saturated gradient a little bit more, this is result:
selection_540

For me it looks fine :)

palob:
Aye, you're getting there.
Alignment is still a bit off + gradient must be from Y position 1 to 47 and 10L lighter at the top.
Btw, in case you don't know you can toggle the pixel grid with the ``#`` in Inkscape.

work

The size of the both red paths (you can merge those if you mant) only needs to be 30x14.
Then duplicate that path, move it 1px down and 1px to the right, colour it black with 10% opacity to have the shadow correctly sized and aligned to the pixel grid.

Jacajack:
Well, I aligned only the bottom part. Should I rescale icon to align both, top and bottom?

palob:
Yes, all horizontal and vertical lines have to be aligned to the pixel grid. Otherwise those lines appear blurry.

Jacajack:
Ok, I'm working on it now, sorry for bothering You :P

palob:
No problem at all, always good to see new contributors :)

Jacajack:
I rescaled symbol to match the grid and set correct gradient end-points.
Visually icon stayed almost the same:
image

palob:
It's correctly aligned :+1:  Size 28x13 now. Maybe that looks even better. Let's see what's @Foggalong's view

Foggalong:
@Jacajack I like the icon, but the gradient is still messed up. It should be only the L (in HSL) value that changes by a value of 10, everything else should remain constant. The gradient nodes should also be aligned to grid and go all the way from the top of the base plate to the bottom in the center.

Jacajack:
@Foggalong Sure, I'll fix that. :)
The weird thing is, that Inkscape won't let me change only L value in gradients, and changes S from 58 to 56 itself, it's pretty annoying. Anyways, I'll try to sort everything out today's evening.

By the way, I think my gradient goes from the bottom of the baseplate to the top, doesn't it?

Foggalong:
Not at the moment no
screenshot_2016-04-21_22-07-20

Jacajack:
That's weird, I can see gradient is at the left side of the icon, and it has proper length 0_0

I will just use new baseplate (unfortunately tomorrow, I had less time than I expected today (sorry!)). And I still have problem with HSL on two different PC's and after updating Inkscape, even when changing other icons' backgrounds.

palob:
@Foggalong That +/- 2 thing is just due to HSL/RGB conversion.

Jacajack:
@Foggalong I used new baseplate in icon, so gradient is perfectly aligned now (I believe I might have broken something in the old one before).

@palob HSL/RGB conversion was first thing that came to my mind actually. In this case I believe I can leave it, as it is, right?

By the way, again, I'm sorry for bothering you that long with just one icon :P

Foggalong:
@Jacajack No problem, thanks for the contribution! :+1:

PR #2869, opened 2016-02-25
Icon for Google Flights

Jacajack:
I created icon for Google Flights web app using information given in [this](https://github.com/numixproject/numix-icon-theme-circle/issues/2862) request.

This is how icon looks:
selection_533

I hope it's fine :)


palob:
Can you recreate the airplane and make it symmetrical?
You're using a flat colour on the baseplate here whereas we use a slight 10L vertical gradient.

Just take the baseplate from an existing icon, click on the handles (circle and square) on the left side to edit the gradient's stop colours.
gradient

Jacajack:
I recreated plane by drawing one half and then mirroring it to make it symmetrical.
I also added background gradient, It must have got lost in time, because I modified this baseplate several times in the past.
Also, plane center is now aligned to grid.

This is result:
selection_541

What do You think about cockpit's shape? Is it better when It's sharper?

Jacajack:
I accidentally closed pull request, sorry. Reopening it.

palob:
Yeah, I think a pointier nose and a stright front fuselage might look better.

Jacajack:
Which shape do You prefer? 
selection_545

I think I would go for the first one, but I'd like to know before I align it and create shadow

palob:
I'd say the second but it's not that crucial.

Jacajack:
I used 2nd symbol then. This is how it looks:
selection_547

It doesn't differ much really, but I changed it.

PR #2886, opened 2016-03-11
gitlab.svg + stremio.svg + github.svg

natanfelles:
It can be used for various purposes.

Especially if you want to disable GitLab automatic startup with:
`systemctl disable gitlab-runsvdir.service`

Then, create a menu item with the following command:
`gnome-terminal -x bash -c "sudo service gitlab-runsvdir start; sudo gitlab-ctl start; exit"`

Disable the automatic startup can be useful if you want to save about 700 MB of RAM.


Foggalong:
@natanfelles Please split this up into one icon per pull request, it makes the review process much quicker. As a note - missing in all your icons is the drop shadow which should be 1px to the right and 1px down of the symbol

PR #2887, opened 2016-03-13
Icon for Calligra Gemini

palob:
https://www.calligra.org/gemini/
QML touch interface for Calligra Words/Stage
calligragemini


palob:
Made the symbol a bit smaller, I think it looks better.
calligragemini

Foggalong:
I can't decide if it needs shifting right slightly to counteract the "triangle" illusion

palob:
It does indeed, it wasn't centered, I had shifted it to the left but 2px was too much with the smaller symbol.
calligragemini

Foggalong:
Much better :+1:

PR #2909, opened 2016-03-25
Icon for PainTown

palob:
http://paintown.org/#/home

Original:
paintown-ori
Pushed:
paintown
Alt:
paintown1


Foggalong:
I'm usually against gradients on symbols but that seems to work pretty well. Go with alt

palob:
Alt pushed :+1:

PR #2916, opened 2016-03-26
Redesigned icons for JetBrains apps. Fixes #2730

### LINKED ISSUE #2730: New icons for JetBrains products
JetBrains, the company behind PhpStorm, IntelliJ and many more IDE's, presented their [rebranding](http://blog.jetbrains.com/blog/2015/12/10/the-drive-to-develop/) yesterday, including new logo's and icons for all of their applications. It would be great if Numix can update the icons that it is supporting right now.

All assets can be downloaded from https://www.jetbrains.com/company/press/
PhpStorm icons are at https://resources.jetbrains.com/assets/phpstorm/phpstorm_logos.zip

phpstorm_400x400_twitter_logo_white

janoj:
New icons for JetBrains apps reflecting their rebranding.

jetbrains-redesign


palob:
Design guidelines aside it looks like you used path clipping. You need to get rid of this otherwise the icons won't show up correctly, especially in Qt-based DEs and apps.

janoj:
Clip paths removed now, as well as some unused gradients.

Regarding guidelines, are the multi-colored background gradients the problem?

palob:
I think so. There's just no other icon which looks like those. Colours might be a bit too garish/saturated. Any thoughts @Foggalong ?
The letters should be coloured L249 instead of pure white and the black could be a bit more greyish (L>=45).

Foggalong:
Yeah, I like the design but can't decide if it's diverging form the guidelines too much to include.

janoj:
A total (well, almost...) step-back in the gradient design can be something like this:

jetbrains_alt

It still does not quite follow the guidelines, but it would be nice to preserve at least some of the variability of the original color palette.

palob:
Looks better imo :+1:

Foggalong:
So much better. Push that new gradient design and after some more verification I'd be happy to merge.

janoj:
Icons with new gradients pushed.

pedroadame:
Nice icons @janoj using them until this PR gets merged

PR #2919, opened 2016-03-29
Letter fix for layout.svg

palob:
Letters weren't converted to path and shadow was a different font. Moved symbol up 1px.


Foggalong:
This is showing up as both green in GitHub preview for me?

palob:
Oops, I had merged the path.

PR #2926, opened 2016-03-31
Icon for Wyrmsun

palob:
https://andrettin.github.io/

wyrmsum-numix


Foggalong:
I'm not sure. The icon is really nice (gradient in particular) but the gradients on the symbol (even though they make it more accurate physically) diverge from the guidelines a bit too much.

palob:
That's how it looks flat:
wyrmsum2

Foggalong:
Ach, that looks nowhere near as good. I'll think about this for a bit see if there's something between which works.

Foggalong:
@palob slightly off topic, but do you fancy joining the Numix developer chat on Telegram? It'd be useful for discussing big picture development strategy which up to now has been clogging up GitHub

ChuckDaniels87:
What about flat shadows?

palob:
@Foggalong Yay, thanks for the invitation.

@ChuckDaniels87 Flat colours on the symbol, like this?
wyrmsun0

ChuckDaniels87:
@palob Yeah, really nice. :wink:

palob:
Pushed!

PR #2928, opened 2016-04-01
Icon for Atomic Game Engine

palob:
https://github.com/AtomicGameEngine/AtomicGameEngine

Original:
atomicgameengine-ori

Pushed:
atomicgameengine


Foggalong:
What does this look like with the symbol rotated 90 degrees so the alignment matches that of the original icon?

palob:
atomicgameengine2
Shall I push?

PR #2984, opened 2016-04-17
Symlink for gscreenshot

palob:
https://github.com/thenaterhood/gscreenshot

### DIFF
--- a/Numix-Circle/48x48/apps/gscreenshot.svg
+++ b/Numix-Circle/48x48/apps/gscreenshot.svg
@@ -1,0 +1,1 @@
+accessories-camera.svg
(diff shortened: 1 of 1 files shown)


Foggalong:
Please symlink screenshot tools to the screenshot icon rather than the camera icon. When we address the associated issue about them being non-distinct this will make things easier :)

palob:
As far as I can see that's where other screenshot tools are symlinked against as well. I could change the link to ``hotshots.svg`` for instance.

palob:
Did it.

Foggalong:
Oh damn, I was thinking there was an `accessories-screenshot` but I seem to have been mistaken

palob:
Revert to ``accessories-screenshot``?

PR #2987, opened 2016-04-17
Symlink for ddrescue-gui

palob:
https://launchpad.net/ddrescue-gui


palob:
Which is what I did meanwhile :)

PR #3000, opened 2016-04-19
Icon for bpython and Python icon fixes

palob:
https://github.com/bpython/bpython

Original
bpython-ori

Pushed
bpython

Alignment fixes for Python icons (pycad and ninja are fine):
Before
before
After
after


Foggalong:
Ooo, thanks!

PR #3006, opened 2016-04-21
Icon for CryptoTE

gregorydk:
Fixes https://github.com/numixproject/numix-icon-theme-circle/issues/2829


palob:
Maybe the lock bolt could be a bit darker for better contrast with the baseplate?

palob:
Doesn't seem to be correctly aligned. I'm going to compile a cleanup PR for the recent additions in the next day so I can fix that.

Foggalong:
Just as a heads up, Circle is hopefully going to be switched over to the Numix Core system over the coming days so the new format will make any old pull requests invalid (why I was sorting through the old pulls).

PR #3007, opened 2016-04-21
Icon for Bitwig Studio

really-need-an-api-key:
An icon for Bitwig Studio.
screenshot from 2016-04-21 15-17-40
Can also provide Square version, if you want.


palob:
The circles aren't displayed properly because they aren't aligned to the pixel grid.


grid

Does it look any good with 3x3 circles aligned to the grid and 2px gaps in between?

really-need-an-api-key:
It looks fine, I guess, I think I just felt it too wide because I was used to seeing it in that particular layout.
bitwig-studio

palob:
Symbol's fine now but there's something wrong with the baseplate gradient. It should be 10L lighter on the top while the saturation value remains constant.

really-need-an-api-key:
Done.
bitwig-studio

PR #3012, opened 2016-04-23
Icon for GNOME News

### LINKED ISSUE #2902: Gnome News
Gnome news is a new RSS reader application for Gnome 
The application is still not released yet (won't be released until 3.22 i guess)
Icon name : `Icon=gnome-news`
Default icon: 
gnome-news
Maybe using Caffeine icon as a base one, and just use the orange color(RSS logo) instead of the current one?

palob:
Fixes #2902


gregorydk:
Using words in icons is not recommended because of localization.

palob:
*News* might be a word which is understood world-wide.

PR #3024, opened 2016-04-25
Sunvox

really-need-an-api-key:
Icon for Sunvox, a modular music digital audio workstation
screenshot from 2016-04-25 17-37-04
.


palob:
Hi, thanks for contributing. The symbol colour should be max. L249 instead of pure white.

Did you use ``svgcleaner-cli`` or ``scour``? 
Your icon's baseplate gradient can't be edited so you need to tweak some options there.

really-need-an-api-key:
Changed the colour to L249.
sunvox svg

I'm not sure I understand the nature of the problem with the gradient, though, because I can edit it after opening the file with no problem.
screenshot from 2016-04-25 23-07-28

Used Inkscape's save as optimized, which I think does use `scour`.
screenshot from 2016-04-25 23-21-38
screenshot from 2016-04-25 23-21-55
screenshot from 2016-04-25 23-22-07

Comparing the files, the only difference seems to be the conversion from css style to attributes, so I went ahead and copied it over from another icon, but I'm unsure if that will fix the issue.

palob:
After you've saved it it will still look like before.

``Convert CSS attributes to XML attributes`` is the toxic option.

really-need-an-api-key:
I'm afraid I can't reproduce that behavior, maybe it's a combination of system/inkscape ver.
Replaced the attrs regardless.

PR #3082, opened 2016-05-09
Symlinks for GNOME Photos and Totem

palob:
Looks like finally all GNOME apps will undergo the ``org.gnome`` treatment.

From what I heard it is neccessary for setting up xdg-bundles for some reason.


palob:
Hold on, I'll add some more.

PR #3088, opened 2016-05-09
Icon for Avogadro

palob:
http://avogadro.cc/wiki/Main_Page

avogadro-icon

avogadro-numix


Foggalong:
Could we possibly use a darker orange fir the background?

palob:
Pushed
avogadro-numix

Foggalong:
Awesome :+1:

PR #3096, opened 2016-05-10
Icon for ISPConfig

natanfelles:


palob:
Symbol isn't properly aligned to pixel grid and the baseplate needs a gradient.
align

PR #3097, opened 2016-05-10
Icon for Nginx

natanfelles:


palob:
alignment issues again and unneeded nodes.
align2

Furthermore, could you try coloured baseplates for your submissions?

PR #3098, opened 2016-05-10
Icon for Stremio

natanfelles:


palob:
Could you recreate the ``3`` using simpler shapes. It's quite irregular now.

Foggalong:
The baseplate has strokes in it. I have no idea how something like that even happens.

PR #3108, opened 2016-05-11
Icon for youdao-dict

Revolution1:
icon for the translation and dictionary app [youdao-dict](http://cidian.youdao.com/index-linux.html)


Revolution1:
origyd
youdao-dict

Foggalong:
Thanks for the contribution, but this icon does not meet our design guidelines. Please use one of the other icons as a template so that you have the right shadows, shades, etc and then we can work from there.

PR #3122, opened 2016-05-12
Icon for Datagrip

really-need-an-api-key:
datagrip
An icon for Jetbrains' Datagrip. Tried to strike a balance between the guidelines and what was said about the gradients in PR #2916 
Not sure if the bar in G should stay misaligned to remain centered or should be moved up/down half a pixel, or maybe the text should be resized vertically (but that would break consistency with other JetBrains icons)?


palob:
How does it look properly aligned? Did you try?

really-need-an-api-key:
I did indeed, but the 'G' ended up stretched and slightly wonky. The best I could do was to re-create the glyph with slightly different proportions, and I think it looks clearer now, centered or not. What do you think?
datagrip-1 datagrip-2

Foggalong:
The one which is aligned to grid looks better, even if it's a little different proportion wise

really-need-an-api-key:
Pushed new changes, then.

PR #3169, opened 2016-05-19
Icon for fs-uae

gregorydk:
Fixes https://github.com/numixproject/numix-icon-theme-circle/issues/2913

### DIFF
--- a/Numix-Circle/48x48/apps/fs-uae.svg
+++ b/Numix-Circle/48x48/apps/fs-uae.svg
@@ -1,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg id="svg2" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48" version="1.1" xmlns:cc="http://creativecommons.org/ns#" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:dc="http://purl.org/dc/elements/1.1/">
+ <metadata id="metadata84">
+  <rdf:RDF>
+   <cc:Work rdf:about="">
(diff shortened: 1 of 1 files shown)


palob:
How does it look with only one check and the letters above it?

gregorydk:
If I do that it's not the logo any more, as seen on https://fs-uae.net/

palob:
The way it is however it clashes with the guidelines w.r.t. both size and gradients on the symbol.

Is it recognisable at smaller sizes?

gregorydk:
Made the logo smaller, it was indeed a bit big (from 32px to 24px width).

palob:
Now the "S" isn't aligned anymore.
misaligned

Moreover there are unneeded nodes which oddly enough even show up in the preview.
preview

Foggalong:
Just for future reference, the colours used in the gradient were also too brash, there are a few loose nodes in the tick symbol, and the text still wasn't aligned to grid. I'm going to merge it and tidy it up because I need to get rid of all the pulls for something we're doing :)

gregorydk:
I know about the colours, but they're the colours from the original logo and didn't want to change that. About the nodes, no idea which ones, I checked before.

Foggalong:
@gregorydk The nodes were in the tick part, usually doubled up on top of one another and hidden away. If you can modify the colours while not straying too far it's worth it to keep a nicer icon. In this case though, [the original Amiga logo](https://upload.wikimedia.org/wikipedia/en/a/a7/Amiga_Logo_1985.svg) provided some softer colours :)

gregorydk:
Ah, I wasn't aware of that logo. I based it on the [fs-uae logo](https://plus.google.com/photos/116108691672931109163/albums/profile/5749218744563325170). I'll see what I can do.

palob:
@Foggalong already chaged the colours accordingly :)

PR #3178, opened 2016-05-25
Add netease-cloud-music's Icon

caoli5288:
Here is the netease-cloud-music's Icon, A online music app from china. 
[website](http://music.163.com)
[debfile](http://s1.music.126.net/download/pc/netease-cloud-music_0.9.0-1_amd64.deb)


palob:
Can you change the symbol size to W 26px H 28px and change the symbol colour to ``f9f9f9ff`` (L249)?

When at the new size, ideally the stroke would be a bit slimmer - exactly 3px:
netease

I've added the shadow in the screenshot: duplicate the symbol, move it 1px down and to the right respectively. Our symbol shadows are coloured black with 10% opacity.

caoli5288:
@palob I think this is a flat icon, do not need shadows. :-)

palob:
Every Numix-Circle icon does have a shadow, take a look please.

caoli5288:
You are right, I will add shadow later.
It may be inappropriate to change size because I got this symbol from origin icon.

PR #3190, opened 2016-05-27
Icon for Xware Desktop

YaoC:
https://github.com/Xinkai/XwareDesktop


Foggalong:
I'm not entirely sure what simplification occurred here, but it's still way too complicated for Numix. Just take a look at some of the other icons - gradients should be kept to a bare minimum, not used throughout the symbol. You also need to keep the shadows consistent with how they're done in other icons.

Issue #2963, opened 2016-04-13
Icon for Franz


brncavalcante:
Icon = Franz
![logofranz](https://cloud.githubusercontent.com/assets/18425086/14506630/25edba4a-0195-11e6-92fc-99bf55e396ea.png)
Thanks :)

iliaskomp:
+1

Foggalong:
Please use GitHub +1s to do this kind of thing, otherwise it's just clogging up the issues tracker. I'll be removing both this comment and yours by the end of the day.

PR #3199, opened 2016-06-01
Icon for franz

kinetoskombi:
Hello, I'm new here and want to contribute to this amazing project. I'm currently using Franz, another amazing project that combines a lot of messaging applications on a beatiful uniform interface. So, I've tried to make an icon based on the official one and other icons of the Numix project.

http://meetfranz.com/

original icon
franz

my proposal
franz


palob:
That's a nice design but the symbol is a bit big. Can you give it a shot at 26px symbol width?

Morevover there's some strange unused stuff in the SVG code. You can get rid of that using Inkscape's ``vacuum-defs`` function (cleaning entry in the ``File`` menu).
For some reason when opening your icon in Inkscape the pixel grid is set to invisible in the document options.

kinetoskombi:
Nica, I'll try to fiz it. Thank you for the feedback!

palob:
Jep, that's it. Can you give it a shot with a slightly blue-tinted background?

Please make sure to open the SVG with a text editor and delete the
```
<!-- Created with Inkscape (http://www.inkscape.org/) -->
```
line in the SVG code. For whatever reason some applications can't get over it.

kinetoskombi:
Resolves #2963

Lazza:
I would suggest to keep the original colors, like this (based on @kinetoskombi's work): [franz.svg.txt](https://github.com/numixproject/numix-icon-theme-circle/files/391960/franz.svg.txt)

franz

Foggalong:
@Lazza It's truer to the original, but it's also a bit boring. I don't personally use the app though while you two do, so if you'd prefer the original colours (hopefully after giving this one a go first) then feel free to open an enhancement request or a pull request and we'll make the change :)

PR #3268, opened 2016-07-05
Icon for Variety Slideshow

gregorydk:
Fixes https://github.com/numixproject/numix-icon-theme-circle/issues/2892


palob:
Maybe without the mountain silhouette for better scalability?

gregorydk:
It's probably better to not change the main icon `variety.svg` for consistency.

palob:
Can you strip a 1px outline around the arrow from the mountain path?

I think the arrow could be set apart a bit more.

Foggalong:
I think it's more the colour that's the problem than the mountain being around the arrow. Try experimenting with some different colours and having the arrows shadow above the mountain.

Also, it needs moving a pixel or two to the right to counteract the triangle illusion.

Foggalong:
I'm still not sold on the design. It just looks a bit off.

Foggalong:
Merging this, but will probably change the design at some point.

Issue #3280, opened 2016-07-07
Icon for World of Tanks


codefuhrer:
It doesn't have an official linux version, but it can be launched via Wine without any problems (described on the [official website](http://worldoftanks.eu/en/content/soft/pc_programs/1311/)).

I made this icon for myself, can PR if it's OK .

![1](https://cloud.githubusercontent.com/assets/2178368/16661712/df24b142-447c-11e6-9abc-c3a46c4179d0.png)

Colors taken from their [Android app](https://play.google.com/store/apps/details?id=ru.worldoftanks.mobile).

palob:
Please feel free open a PR, it's hard to do the review just based on the PNG preview.

PR #3283, opened 2016-07-09
Added World of Tanks icon

codefuhrer:
See #3280


palob:
Hi, thanks for the PR.
The symbol is a bit big. It shouldn't exceed a size of 26px X 26px (28x28 for some exceptions or 30x30 for circular symbols).

In order not to appear blurry all horizontal/vertical straight lines need to be aligned to the pixel grid (toggled with ``#`` in Inkscape).
blurry

codefuhrer:
Ok, will do.

This icon's bottom part is optically lighter, than the top. Should it be exactly 26x26px anyway?

codefuhrer:
Hm. Some parts are too far from any grid intersection:

Screenshot

If I just align everything to the grid, it'll look like BS. What should I do?

palob:
You need to find a size symbol size where it works. I'm sure you will :) I've done a lot of icons and there's always been a way to get the alignment right without sacrificing too much of the symbol characteristics.

It can take a few shots sometimes.

Foggalong:
@codefuhrer Do you have plans to update this pull request?

codefuhrer:
@Foggalong Not really, cannot find the right position. 
I don't mind if you want to close it.

Foggalong:
No problem, since it's your first icon and we've done all this messing around with repo structures I'll merge it and fix it up :+1:

codefuhrer:
Cool, looks much better after your fixes :hushed:
Thanks!

PR #3309, opened 2016-07-24
Symlink for filerunner

palob:
https://sourceforge.net/projects/filerunner/


Foggalong:
screenshot_2016-08-01_17-21-45
Strong contender here for the worst original icon award.

PR #3345, opened 2016-08-02
linuxdcpp - duplicate to symlink

palob:


Foggalong:
I'm confused why these are duplicates? The icons do have distinct designs

palob:
``linuxdcpp`` is just slightly distinct because it's the old misaligned design.

Foggalong:
Oh right, they're actually icons for the same app then?

palob:
From what I can see it's those two


https://launchpad.net/linuxdcpp
http://dcplusplus.sourceforge.net/

Foggalong:
In which case yes, keep the distinct icons over a symlink.

palob:
Well, the design isn't distinct. It's just that I did some alignment work for ``DC++.svg`` which ``linuxdcpp.svg`` didn't receive. Prior to that it was aproper duplicate.

Foggalong:
One is [orange](https://github.com/numixproject/circle-core/blob/master/icons/48/linuxdcpp.svg) and one is [green](https://github.com/numixproject/circle-core/blob/master/icons/48/DC++.svg) though?

palob:
Oh, alright, will recolour DC++.svg then :)

PR #3355, opened 2016-08-04
Removed path artifacts from templates

palob:
nodes


Foggalong:
Not sure how I managed to get paths in there. As you've probably guessed, once Square is updated and a considerable amount of the open issues resolved the plan is to do what should have been done from the off and flesh out the Circle icons in more than just 48px.

Issue #2183, opened 2015-05-27
More Distributor Logos


Foggalong:
I use our distributor logos for users flairs over in [/r/unixporn](https://www.reddit.com/r/unixporn). For that reason I'll probably be adding these new icons in the near future. Note that there are some flairs that aren't currently logos and vis versa so I'll make things equal before I start off on this.
- [x] Aix
- [x] Android
- [x] Antergos
- [x] Antix
- [x] Arch
- [x] Arya
- [x] Archbang
- [x] Backbox
- [x] Bodhi
- [x] BSD
- [x] CentOS
- [x] Chakra
- [x] Chalet OS
- [x] ConnochaetOS
- ~~[ ] CrunchBang~~
- [x] Crux
- [x] Cub Linux
- [x] Debian
- [x] Deepin
- [x] DragonFly BSD
- [x] elementary OS
- [x] Endless OS
- [x] Exherbo
- [x] Fedora
- [x] FreeBSD
- [x] Funtoo
- [x] Gentoo
- [x] GhostBSD
- [x] GNU
- [x] GuixSD
- [x] HandyLinux
- [x] Illumos
- [x] Kali
- [x] KaOS
- [x] KDE Neon
- [x] Korora
- [x] Kubuntu
- [x] LFS
- [x] Lite
- [x] Linux
- [x] LMDE
- [x] Lubuntu
- [x] LXLE
- [x] Mageia
- [x] Manjaro
- [x] Maui
- [x] Mint
- [x] MS-DOS
- [x] MX Linux
- [x] NetBSD
- [x] NixOS
- [x] OpenBSD
- [x] OpenIndiana
- [x] OpenMandriva
- ~~[ ] Open Solaris~~
- [x] openSUSE
- [x] OSX
- [x] Parabola
- [x] Parsix
- [x] PC-BSD
- [x] PCLinuxOS
- [x] Peppermint
- ~~[ ] PiBang~~
- [x] Plan 9
- [x] Puppy
- [x] Qubes
- [x] Raspbian
- [x] Red Hat
- [x] Rosa
- [x] Sabayon
- [x] Salix
- [x] Scientific Linux
- [x] Siduction
- ~~[ ] Simplicity~~
- [x] Slackware
- [x] SmartOS
- [x] Solaris (Oracle)
- [x] Solaris (Sun)
- [x] Solus
- [x] SparkyLinux
- [x] Tails
- [x] Tanglu
- [x] Trisquel
- [x] TrueOS
[...]

PR #3359, opened 2016-08-06
Some more distributor icons

palob:
More for #2183 

distros4


Foggalong:
Can we make the symbols for solus and netrunner slightly smaller?

palob:
distro222

PR #3361, opened 2016-08-06
Next cleanup/alignment spree

palob:
Removed ``plague2.svg`` which I had introduced accidentally earlier.


Foggalong:
All good apart from pxsc and qtsixa. The symbols are based on the playstation controller layout and the button arrangement in pxsc drifts away from that a bit

palob:
The new ``pcsx-icon`` is ``epsxe`` recoloured and doesn't differ greatly from how it looked before?

Should I base ``pcsx-icon`` on ``qtsixa`` or just arrange the buttons on the right hand in cross manner?

Foggalong:
Just arrange the rhs buttons for pcsx in a cross for now and we'll possibly come back to this in future

palob:
:+1:

Foggalong:
:+1:

PR #3366, opened 2016-08-09
Enpass icon proposal

cpriego:
I created this Enpass proposal icon for the one included in the pack doesn't resemble the original application. Hope it can be included unless there is some copyright issue.


palob:
Hi, thanks for your contribution. When I open your icon in Inkscape it's broken. Please don't use any filters for your icon, notably no blur as it is slow to render and doesn't fit the Numix style.

palob:
Now it shows up correctly.

I think to be more consistent with existing icons it's best to go without the outer white ring.

The white circle should by sized exactly 30px X 30 px placed at X/Y 9.

The keyhole should be aligned with the pixel grid (press ``#`` to toggle it in Inkscape):
keyhole

Afterwards it won't look blurry anymore.

cpriego:
Ok, I made the corrections, sorry about the double commit, I made a mistake.

This is how the icons looks now.
enpass-numix

palob:
Horizontal lines are aligned :+1: . Now you need to apply the same treatment to all vertical borders in order to get something like the yellowish shape below:

keyhole2

cpriego:
Hahahaha, coming close to the end result, The keyhole is now adjusted both horizontally and vertically.
enpass-numix

Foggalong:
I'm not sure on this. The current icon is much nicer design wise and does keep the iconography of the original, but it doesn't stick to it design wise which is something we like to do.

Foggalong:
Really sorry to have waisted your time on this but we're gonna stick with the current design. It still keeps the keyhole iconography but the style that Enpass went with is a bit to plain for my liking.

PR #3371, opened 2016-08-10
Icon for Neverwinter Nights

Ewaryst:
A BioWare RPG game ported to Linux in June 2003


palob:
Hi, thanks for your contribution.

Drop shadow colour should be black with 10% opacity.
The symbol is a bit big. Can you give it a shot with 28px width and try to align the tears to the pixel grid?

palob:
That's a better size. The symbol should be moved down 1px and maybe the baseplate could be a bit brighter for improved contrast.

The symbol itself might be good to go the way it is yet there are some rough shapes (due to a few unneeded rough nodes) and slightly blurry parts (due to misaligned vertical/horizontal shapes/lines).

If you are in a mood for tinkering you can use the current symbol as a stencil for recreating it using  properly aligned objects like circles, paths, rectangles and combining/subtracting the with the actions from Inkscape's ``Path`` menu.

Here's a screenshot showing what I mean (only changed the background colour for better pixel grid visibility):

work

Ewaryst:
Okay, I'd be happy to play around. Thanks for advice

palob:
One more hint in this regard: If you go this way and use the symbol (or any original PNG/SVG ... icon) as stencil make sure the object you want to recreate has the right size before you start.

Otherwise in case you have to resize the new symbol afterwards you will probably lose alignment again.

Ewaryst:
I've recreated it using this method and everything looks well aligned except for 'horns' (they're a bit hard to create in a non-messy way :/)
scrn
I'll assemble it now and see how it looks

Foggalong:
Does this icon use clipping? It's not displaying correctly in the GitHub preview. Also I think the symbol should probably be moved down a pixel or two.

palob:
Yep , it's really simpler now :+1: 

There's a tiny jag left on each side, just delete the node I selected in the screenshot and retract the handles of the adjacent remaining nodes.

jag1

jag2

Ewaryst:
Sorry for delay.
Thank you for your guidence, I'll try to pay more attention to details from now on.

As to icon not displaying correctly on GitHub preview, I believe it had something to do with some invisible leftovers in previous commits, but not sure. I think it should be okay now.

PR #3381, opened 2016-08-17
Fixes #3347

### LINKED ISSUE #3347: Chrome Web Store icon missing
This is the last missing one on my system I promise! :) Unfortunately I cannot provide a link for the chrome webstor webapp since its its the store itself missing the icon. The application ID and icon ID is: **chrome-ahfgeienlihckogmohjhadlkjgocpleb-Default**

The original app icon looks like this:
chrome internetes aruhaz

Based on your other software center icons something like this would look great I think:
softwarecenter

Thanks again for the great work you guys do!

palob:
Fixes #3347 

chrome-ahfgeienlihckogmohjhadlkjgocpleb-default


Foggalong:
The colours of the yellow segment and the bag are a little too similar. Possibly try a different bag colour?

palob:
Desaturated it a bit.
chrome-ahfgeienlihckogmohjhadlkjgocpleb-default

Foggalong:
That's a bit better, but still not great. Possibly a slightly darker yellow and lighter bag at the same time?

palob:
chrome-ahfgeienlihckogmohjhadlkjgocpleb-default

palob:
Pushed this:
chrome-ahfgeienlihckogmohjhadlkjgocpleb-default

PR #3386, opened 2016-08-21
Icon for System Workbench for STM32

Jacajack:
I made an icon for for *System Workbench for STM32* - an Eclipse based IDE for developing software for STM32 ARM cores.

So, this is the icon itself:
selection_803

Also, the butterfly is not symmetrical, because the original one isn't either. Of course, I can change that if that's a problem.

I tried to follow guideline, and I'm willing to fix anything, just please let me know what's wrong :smile:


palob:
Hi, thanks for your contribution. Can you please change the symbol colour to ``f9f9f9ff``?

Jacajack:
Sure, I completely forgot about that, but there it is. If I may ask, is there any specific reason we shouldn't use white color in symbols?

palob:
Numix Circle icons are designed without using eye-jarring/fully saturated colours. Same for white and black where there's saturation per se.
There's no specific reason which would generally rule out pure white for icon design, at the end of the day it's an arbitrary decision for a coherent design. For the Numix base theme for instance we use full-blast white.

However if you want to have darker/brighter structures on a solid colour surface (like the bezel shadow on the Numix Circle baseplate) you can't go with pure white/black obviously.

palob:
Sure, if you're serious about this it's probably best to discuss this in an https://github.com/numixproject/numix-icon-theme issue.

This might be related as well https://github.com/numixproject/numix-folders/issues/78.

palob:
Hi, now I've seen the icon along with others I the symbol looks slightly too big.

Can you try a size 26x24 (w x h) located at X10 Y13 ?

Jacajack:
@palob Hi, sure, there you go :)

Issue #3270, opened 2016-07-06
Icon for RationalPlan


btd1337:
A Numix-circle-icon for RationalPlan is missing.

This software cointains 3 versions: Project, Single, Multi.

![captura de tela de 2016-07-06 01-17-08](https://cloud.githubusercontent.com/assets/11943860/16607449/658fc788-4317-11e6-979e-d38b9a49747a.png)

**The background color is related to your version.**

![captura de tela de 2016-07-06 01-20-23](https://cloud.githubusercontent.com/assets/11943860/16607485/d8d3efb2-4317-11e6-8144-b467bce6e917.png)

Icon=rationalplan-single
Icon=rationalplan-multi
Icon=rationalplan-project

Website Project: http://www.rationalplan.com/index.php

Jacajack:
@Foggalong I think this can be closed :)

Foggalong:
Indeed it can :)

PR #3394, opened 2016-08-26
Icons for RationalPlan

Jacajack:
Hi everybody!
I made icons for RationalPlan as requested in #3270 

selection_837

I hope everything is ok :)

### DIFF
--- a/Numix-Circle/48x48/apps/rationalplan-multi.svg
+++ b/Numix-Circle/48x48/apps/rationalplan-multi.svg
@@ -14,7 +14,7 @@
    inkscape:version="0.91 r"
    width="100%"
    height="100%"
-   sodipodi:docname="rationalplan-project.svg">
+   sodipodi:docname="rationalplan-multi.svg">
   <metadata
      id="metadata67">
     <rdf:RDF>
@@ -23,7 +23,7 @@
         <dc:format>image/svg+xml</dc:format>
         <dc:type
            rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
-        <dc:title></dc:title>
+        <dc:title />
       </cc:Work>
     </rdf:RDF>
   </metadata>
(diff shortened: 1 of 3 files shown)


palob:
It's a little bit blurry in some places. Con you find a configuration where the pixel grid alignment is better to get a really crips symbol. For instance the shape I've selected below will get crisper by just moving it some 0.4 px down.

selection

The lowermost path gets crisper by changing the height to 4 and moving it to an integer Y position.
The head is crisper with integer dimensions and x/y positioning.

Jacajack:
@palob Thanks for advice, is that better now?

I rounded width, height, X and Y of each path to integers, so it should be better (except from arm part(?), because I assumed it's ok).
By the way, I thought that only straight horizontal and vertical lines should be aligned with the grid, so why does that blur happen?

palob:
:+1: Better now.

Proper alignment is important for curved (but almost vertical/horizontal) and ellipsoid shapes as well. 

Basically in order to get an icon as crisp as possible you need to get as many of the"pixel grid squares" in Inkscape (each representing a single pixel on the screen) to be filled with only one colour (or predominantly one colour). 
If the edge between two differently coloured shapes runs through one of those "squares" so that there are two different colours in it the pixel on screen will be displayed with an interpolated colour (interpolation based on area fraction of the colours).
Now if an object isn't properly aligned to the pixel grid (meaning the "colour edge" isn't congruent with pixel boundaries) there will be a line of interpolated colouring around this object which is the root for blurriness.

Jacajack:
Oh... That makes sense...
Now I don't know how I could not figure that out myself.

I will try to take more care of it next time
Thanks  :)

palob:
One "warning" in this regard: Inkscape grid = pxel grid isn't cast in stone.

It might not be set up that way with every SVG file out there.

Actually Inkscape lets you set the grid lines' intervall and direction freely in the document options.

Jacajack:
I understand.
I think I'm going to make checklist from all the mistakes I made, not to bother you too much, when I contribute in future.
Thank you for explanation once more :)

palob:
I'll probably add more detailed, illustrated step-by-step instructions as to how to create a Numix Circle icon to the wiki some time.
Thanks you plan on continuing to contribute :)

PR #3399, opened 2016-08-31
nth alignment

palob:
@Foggalong Fixed folder structure for files form #3394 and #3386 (and svgcleaned). There were in a subdirectory ``Numix-Circle/48x48/apps``.


Foggalong:
I totally missed they were in the old directory, thanks for catching that :+1:

PR #3401, opened 2016-09-01
Improved icon for CuteCom

Jacajack:
Hi!

Normally CuteCom uses same icon as terminal, which causes a lot of confusion, at least in my case, because 100% of the time I click the opposite of what I want...

image

This is the icon I made:
selection_867

Only thing that worries me is that desktop file wants to use `openterm` icon, which is linked to the terminal icon.

This is the content of `/usr/share/applications/kde/cutecom.desktop`
```
[Desktop Entry]
Encoding=UTF-8
Name=CuteCom
MimeType=
GenericName=Serial Terminal
Exec=cutecom
Icon=openterm
Path=
Type=Application
Terminal=0
X-KDE-StartupNotify=false
```

Does it count as a hardcoded icon? If so, what can I do with it?


Foggalong:
What does CuteCom do?

palob:
@Jacajack Please do a ``git pull`` to update your fork. You are still pushing your new icon inside the old directory structure.

Jacajack:
@Foggalong, it's a serial terminal, this is how it looks:
cutecom_880

Icon represents DB9 connector, which is used for RS232 serial communication standard.

@palob Whoops... I didn't notice that you changed something. Let me try to fix it

palob:
Is there any app named *openterm*?

Jacajack:
@palob Uh... That's exactly what I was thinking about. I think Google search for `openterm` doesn't return any useful information. Maybe it's just old name for CuteCom?

Foggalong:
[Found this](http://forum.nsu.ru/viewtopic.php?t=567). It appears to be something to do with konsole, and our icon originally came from [the faenza commit](https://github.com/numixproject/circle-core/commits/ef3d5f17aa5c45fc3ce06d0c8dc58adc7b4b3559/Numix-Circle/48x48/apps/openterm.svg).

Jacajack:
@Foggalong I asked about that. To be honest, posts on that forum are 14 years old, so I'm not sure if they are still up to date.

Foggalong:
@Jacajack Honestly no idea. It's most probably that Konsole has long since changed it's icon line from openterm. It wouldn't be the first time that we've had to change something because we picked up an archaic icon from Faenza. However it's also inadvisable that CuteCom pick an icon name the same as that of another program, however old the latter may be.

Jacajack:
Alright, since CuteCom now refers to `cutecom` icon (https://github.com/neundorf/CuteCom/pull/12 and https://github.com/neundorf/CuteCom/issues/11), I changed icon name to `cutecom.svg`.

Jacajack:
@Foggalong, yep, that's exactly what it was for. I rescaled the background part as you advised me to do and aligned the circles. Honesty, the version you presented above (with only circles aligned) also looks good, and at first I didn't even realize that they were not centered along Y axis.

PR #3442, opened 2016-10-09
Icon for Peek

palob:
https://github.com/phw/peek
peek
peek-numix

Alts:
peek0
peek2


Foggalong:
@palob This has two copies of the icon in, one in the old location and one in the new :P

PR #3444, opened 2016-10-09
Icon for fgmk

palob:
https://github.com/ericoporto/fgmk

icontiny
fgmk-numix


Foggalong:
@palob This looks great, but we also need the `data.json` lines like you did [here](https://github.com/numixproject/numix-core/pull/3465/files)

PR #3445, opened 2016-10-10
Icon for Jetbrains Rider

### LINKED ISSUE #3226: Icon for JetBrains Project Rider C# IDE
This IDE is currently still in development so that is probably why it currently doesn't have an icon
Also unfortunately, like all other JetBrains IDEs (IntelliJ, PyCharm etc..), its icon is hardcoded. However I really like the new ones you have made for both IntelliJ and PyCharm so one for this would be a nice addition

rider_128

This is the current Project Rider Icon. The name of the IDE is probably subject to change as well due to it being in early development

Desktop file: `jetbrains-rider.desktop`
Icon line: `/opt/JetBrains Project Rider/bin/Rider_128.png`

However as the image path is not set. The /opt/JetBrains Project Rider/ is relative to where the user has chosen to install the application. So the **bin/Rider_128.png** would be the important part.

I think that Issue #2730 covers JetBrains products however this one didn't get done because they are still developing it

really-need-an-api-key:
Fixes #3226. Application is still in early access, but the name hasn't changed and hopefully won't. I was bored anyway.
That gradient might be a little too much?
rider 48 svg
rider 192 svg


Foggalong:
@really-need-an-api-key Can you also add the lines needed for the `data.json` file?

PR #3446, opened 2016-10-10
Icon for relational

palob:
https://github.com/ltworf/relational

relational
relational-numix


Foggalong:
@palob This has two copies of the icon in, one in the old location and one in the new :P

PR #3449, opened 2016-10-13
Remove git dep & add pre-existing repo options

bilelmoussaoui:


Foggalong:
These are both now dealt with because of the repo merge :frowning_face:

PR #3450, opened 2016-10-13
Repo Merge with Numix Core

Foggalong:
The [Numix Core](https://github.com/numixproject/numix-core) repo is merging into circle-core so that all the materials for the building of themes is in one location. It seemed to be causing a lot of confusion as to where issues for certain things should be posted so hopefully this will clear things up a bit.


Foggalong:
I'm kinda stunned. I was expecting this to be horribly broken but it worked with very little effort. Neat :+1:

PR #3466, opened 2016-10-30
Added missing DB entries

palob:
For PRs #3426 #3427 #3429 #3440


Foggalong:
Thanks, my bad on missing those :frowning_face:

PR #3467, opened 2016-10-30
Icon for iQPuzzle

palob:
https://launchpad.net/iqpuzzle

iqpuzzle_64x64
iqpuzzle


Foggalong:
The blue of the middle piece and the background clash a bit too much - possibly an orange for that piece instead?

palob:
Pushed :+1: 
iqpuzzle

PR #3473, opened 2016-11-03
Add bfb icons support

bilelmoussaoui:
- bfb icons will only be generated from 48 sizes (things that we can change in the future) but i've added a condition to prevent having a bfb icon generated for each size which won't be needed...
- I have also added a small message error when the user only downloads the script and data.json file (if the icons folder is not being found) 

In order to add that to data.json, for each icon we can add 
` "bfb" : "launcher_ubuntu"`

### DIFF
--- a/gen.py
+++ b/gen.py
@@ -30,7 +30,10 @@ with open('data.json') as data:
     icons = load(data)
 
 # User selects the theme
-theme, themes = "", listdir("icons")
+try:
+    theme, themes = "", listdir("icons")
+except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


Foggalong (changes requested):
This works great in terms of generating the correct BFB icon, but it does it instead of copying the root icon and creating the symlinks. Instead it should be doing both

bilelmoussaoui:
Fixed that in the latest commit

Foggalong (approved):

Issue #3317, opened 2016-07-26
Icon for avocode


simondubois:
A numic-circle icon for avocode is missing.
The application can be downloaded at : https://avocode.com/download/
The default icon name is `avocode`.
The default icon is provided in the zip file and is attached to this issue.
![avocode](https://cloud.githubusercontent.com/assets/9416595/17127856/10ed4a78-5309-11e6-9ed0-11d20f74f5b0.png)

PR #3476, opened 2016-11-05
Icon for Avocode

Jacajack:
selection_002

Icon for Avocode as requested in https://github.com/numixproject/numix-core/issues/3317

### DIFF
--- a/data.json
+++ b/data.json
@@ -1032,7 +1032,7 @@
             ]
         }
     },
-	"avocode": {
+    "avocode": {
         "linux": {
             "root": "avocode"
         }
--- a/data.json
+++ b/data.json
@@ -1032,6 +1032,11 @@
             ]
         }
     },
+	"avocode": {
+        "linux": {
+            "root": "avocode"
+        }
+    },
     "avogadro": {
         "linux": {
             "root": "avogadro",
--- a/icons/circle/48/avocode.svg
+++ b/icons/circle/48/avocode.svg
@@ -1,0 +1,138 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   viewBox="0 0 48 48"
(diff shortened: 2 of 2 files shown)


Foggalong:
I know you're just sticking to the original icon, but we generally avoid gradients like this in icons because it goes against our design guidelines. In this situation I'd probably recommend using a darker green for the background gradient and then a solid light green for the symbol, taking both colours from the gradient in the original icon if possible.

Jacajack:
Sure, give me a second :)

Jacajack:
imageselection_006

I made two version of the icons with swapped colors (I can tweak them a bit If you like). Which one do you prefer?

Foggalong:
I think definitely no gradients and colourful background is the way to go, but maybe we could tweak the symbol a bit so it resembles an avocado a bit more? Something like this  
screenshot_2016-11-06_15-27-09

If you're not up for that just let me know and I'll do the tweaking :+1:

Jacajack:
selection_999 036

Something like this?

Foggalong:
@Jacajack That, with then a slightly smaller version of the outer egg part coloured yellow layered above the light green. Then add a shadow like the main symbol shadow for the stone against the yellow

Jacajack:
selection_999 042

Yeah, that's a real avocado now. At first I wanted the icon to look more or less like the original one, but I like this design even more. :)

Foggalong:
That's looking great! Push that and then I'll do one last code checkover before merging :)

Jacajack:
If you ever need an egg icon, remember about me :stuck_out_tongue_closed_eyes: 
selection_999 043

Foggalong (changes requested):
Excellent, last few nit picky changes now just based on the SVG content rather than the iconography.
- [ ] Increase the size of the light green egg to 20px by 26px with location (14,11). Our standard symbol size is 24px by 24px so this is a nice compromise. Remember to move & resize the background shadow accordingly.
- [ ] Increase yellow eggs size to 18px by 24px with location (15,12). This just keeps everything grid aligned.
- [ ] Same with the stone, increase to 10px by 13px with location (10,15). Remember to keep the shadow for the stone 1px to the right and 1px down, the one pushed is slightly closer than that.
- [ ] When saving the file, save as "Optimised SVG". This means it takes up less file space in packaging but keeps the exact same design

With those done this is good to go :+1:

data.json
```diff
@@ -1032,6 +1032,11 @@
             ]
         }
     },
+	"avocode": {
```
Foggalong: This also has the thing with the tap instead of a space :stuck_out_tongue:

Foggalong (changes requested):

Jacajack:
untitled window_999

Inkscape throws an error when I'm trying to save as optimised SVG. :confused:
Can I use `svgo` or something like this?

<hr>

_Oh... how nice..._
selection_999 044

Jacajack:
After manually retyping some Python, I fixed that. I hope everything is alright :stuck_out_tongue:

Foggalong:
Yeah, it can be a bit messy with that. Thanks for powering through the errors though :)

Foggalong (approved):

Issue #3315, opened 2016-07-25
Icon for FUSE - ZX Spectrum emulator


sljunkie:
Probably the best 'spekkie' emulator around.

Homepage: http://fuse-emulator.sourceforge.net/

```
$ cat fuse-emulator.desktop 
[Desktop Entry]
Type=Application
Name=Fuse ZX Spectrum Emulator
Comment=Emulator of the 1980s ZX Spectrum home computer
Icon=fuse-emulator
Exec=fuse %f
Terminal=false
Categories=Utility;Emulator;
```

PR #3477, opened 2016-11-06
Icon for Fuse ZX Spectrum Emulator

Jacajack:
image

Icon for Fuse Emulator as requested in https://github.com/numixproject/numix-core/issues/3315.
I'm not sure if shadow located under the rainbow is necessary.

### DIFF
--- a/data.json
+++ b/data.json
@@ -4643,6 +4643,11 @@
             "root": "furiusisomount"
         }
     },
+	"fuse-emulator": {
+        "linux": {
+            "root": "fuse-emulator"
+        }
+    },
     "fusion-icon": {
         "linux": {
             "root": "fusion-icon"
--- a/data.json
+++ b/data.json
@@ -4643,7 +4643,7 @@
             "root": "furiusisomount"
         }
     },
-	"fuse-emulator": {
+    "fuse-emulator": {
         "linux": {
             "root": "fuse-emulator"
         }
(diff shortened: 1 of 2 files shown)


data.json
```diff
@@ -4643,6 +4643,11 @@
             "root": "furiusisomount"
         }
     },
+	"fuse-emulator": {
```
Foggalong: There's on indent too far here. It's not a big deal but it breaks consistency

Foggalong (changes requested):
The design is nice, but it could be made more consistent with our design guidelines by cropping and centring the logo (and possibly reducing the angle toward the vertical too) like it's displayed on this machine.

data.json
Jacajack: Weird... It looks alright in gedit... :confused:

Foggalong: The `data.json` uses spaces rather than tabs, but this line in your commit uses tabs instead. None of the other lines you've committed do though :stuck_out_tongue:

Jacajack:
@Foggalong Again, I have two variations.

selection_008selection_007

I think the first one looks little bit sharper.

Foggalong (approved):

Foggalong (changes requested):
The isolated symbol with the gaps between the lines (option 1) looks much better. Maybe reduce the angle ever so slightly more though? Then you might be able to get away with a slightly larger symbol (up to 26px^2)

Jacajack:
I managed to make the stripes wider and more sloped. Square symbols don't really look good in this case (at least I think so), so I made rectangular ones:

26x19:
selection_999 041

24x17:
selection_999 040

Foggalong:
26x19 looks great! Push that and I'll merge :)

Foggalong (changes requested):
This all looks good to me, just needs saving as an optimised SVG and we're good to go :+1:

Foggalong (approved):

PR #3489, opened 2016-11-08
add onlyoffice symlink

bilelmoussaoui:


data.json
```diff
@@ -857,7 +857,10 @@
     "asc-de": {
         "linux": {
             "root": "asc-de"
-        }
+        },
+        "symlinks" : [
```
Foggalong: This isn't the right formatting to add symlinks :P

Foggalong (changes requested):

data.json
bilelmoussaoui: My bad :P was so tired

bilelmoussaoui:
Fixed

PR #3499, opened 2016-11-18
Add --theme and --platform args

bilelmoussaoui:
This will allow me to build git versions for AUR package users 
capture du 2016-11-18 21-49-35
capture du 2016-11-18 21-50-38

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,11 +14,9 @@ To deal with hardcoded application icons Numix uses the [hardcode-fixer](https:/
 ## Script
 
 ### Dependencies
-The script needs Python 3.x to run. The downloading of the icon resources requires the [GitPython](https://github.com/gitpython-developers/GitPython) module and for [git](https://git-scm.com/) to be installed and accessible via system's PATH. The exporting-to-PNG part of the script currently uses [Cairo](https://cairographics.org/) or [Inkscape](https://inkscape.org/). The OSX packaging needs [libicns](http://icns.sourceforge.net/) for the `png2icns` command.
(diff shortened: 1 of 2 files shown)


Foggalong:
If we're gonna do flags, I'd like to do flags only if no one has any objections.

bilelmoussaoui:
You still can run the script without any flags

Foggalong:
@bil-elmoussaoui Yeah, that's what I mean - ditch the in script prompts so that the way to build is using tags.

bilelmoussaoui:
@Foggalong Well, i thought that keeping the prompts is good for beginners :) but i can do that if you want so :+1:

Foggalong:
@bil-elmoussaoui I see your point, but ideally we don't want beginners building from this repo. They should be using the packaging repo

bilelmoussaoui:
@Foggalong Done, i've also added a condition when the data.json file can not be found

bilelmoussaoui:
I've also removed the GitPython depend form the README as we don't use it anymore, and updated the instructions, it might not be clear enough..

Foggalong:
I get this error regardless of whether flags are used

screenshot_2016-11-23_22-44-54

bilelmoussaoui:
I've already fixed this issue but i forgot to commit the modifications :/ Done

Foggalong:
@bil-elmoussaoui Made some changes to fix a few small issues and then neaten things up. Any more changes to make / alterations to mine?

bilelmoussaoui:
It seems all good for me :+1:

Issue #3481, opened 2016-11-07
100% Coverage for Square


Foggalong:
At the moment not every Circle icon has a Square equivalent. From now on all pull requests will be required to include both Circle and Square version (even if that means one of the Numix team adding them) but for the existing icons this will still need addressing.

This list is to be prioritised over #3482
- [ ] kitematic
- [ ] KKEdit
- [ ] KKEditRoot
- [ ] klatexformula
- [ ] klayout
- [ ] klettres
- [ ] klickety
- [ ] klines
- [ ] kmag
- [ ] kmahjongg
- [ ] kmplot
- [ ] knavalbattle
- [ ] knetattach
- [ ] knetwalk
- [ ] knights
- [ ] knowthelist
- [ ] koala
- [ ] kolf
- [ ] kollision
- [ ] kompare
- [ ] kongdash
- [ ] konquest
- [ ] korora-welcome
- [ ] kreversi
- [ ] kronometer
- [ ] kshisen
- [ ] kshutdown
- [ ] ksirk
- [ ] ksnakeduel
- [ ] kspaceduel
- [ ] ksquares
- [ ] kst
- [ ] kstars
- [ ] ktip
- [ ] ktuberling
- [ ] kturtle
- [ ] kvirc
- [ ] kwave
- [ ] kwordquiz
- [ ] laby
- [ ] labyrinth
- [ ] lantern
- [ ] lash_panel
- [ ] latexdraw
- [ ] layout
- [ ] LeagueOfLegends
- [ ] lenmus
- [ ] librevault
- [ ] lifeograph
- [ ] lime
- [ ] mynotex
- [ ] mytourbook
- [ ] myunity
- [ ] naev
- [ ] nagstamon
- [ ] nasc
- [ ] natronIcon256_linux
- [ ] nautilus-pastebin
- [ ] nautilus-scripts-manager
- [ ] navit
- [ ] neovim
- [ ] nestopia
- [ ] netease-cloud-music
- [ ] netgen
- [ ] nextcloud
- [ ] nfoview
- [ ] nginx
- [ ] nightingale
- [ ] nip2
- [ ] nodejs
- [ ] NoMachine-icon
- [ ] nootka
- [ ] notepadqq
- [ ] nsight
- [ ] numix-folders
- [ ] nutty
- [ ] nvvp
- [ ] nwn
- [ ] nylas
- [ ] obozrenie-short
- [ ] ocrfeeder
- [ ] ogmrip
[...]

PR #3501, opened 2016-11-18
Calligra icons for Square + alignment fixes

palob:
For #3481

calligra

Alignment fixes for existing icons.


andia89:
A question regarding Square Icons: could this be scripted? (A tleast for Icons that involve Symbols) it should be poasible to extract the syMbols and put them on a Square canvas with the Same color as for the Circle icons

palob:
I'm not sure if finding the symbol is such a trivial task. What's grouped differs from symbol to symbol. 

It's probably still helpful even if the script yields a few broken icons.

andia89:
Ah that's easy I guess just keep everyrhing that's not the baseplate. Do we
need to change Symbol sizes/Position etc? I guess i could put something
together.

Vincent Bermel notifications@github.com schrieb am Fr., 18. Nov. 2016,
23:20:

palob:
Symbols need to go 1px up.
Symbol shadows (except for exceptions) need to be shifted 1px to the left.

andia89:
Damn that makes the whole thing a lot harder...

palob:
Isn't replacing the baseplate in Circle icons or working with templates a more elegant way of doing this?

This is what these scripts do:
https://github.com/numixproject/numix-labs/tree/master/Satya/script-o-hacky
https://github.com/numixproject/numix-labs/blob/master/Satya/script-o-mighty/numix-kit

andia89:
Sure but where to we get the symbols from, Position it and move the Shadow?

palob:
If it hasn't been cleaned away the easiest way is to look for the `id` each SVG object has.

The symbol is everything but the baseplate objects.

It's probably best to run `inkscape --vacuum-defs` on all icons first to get a uniformly formatted code base.

Foggalong:
Please move this discussion to #3481 otherwise it's just going to get lost.

PR #3522, opened 2016-12-10
Icons for CINEMA 4D

really-need-an-api-key:
Icon for [CINEMA 4D](https://www.maxon.net/en/products/cinema-4d/overview/), a 3D modelling package.
c4d_logo
screenshot from 2016-12-10 19-39-59

It's actually a Windows program, but works pretty well under Wine.
However, the names of the icons (`9FD2_Commandline.0`, `38F9_CINEMA 4D TeamRender Server.0`, `60EE_CINEMA 4D.0`, `F879_CINEMA 4D TeamRender Client.0`) Wine creates have spaces in them, so they would cause the icon cache to be invalid.
I wasn't sure whether to use them, so didn't and opted for a neutral `cinema4d` filename, but I don't know if that's right.


palob:
Symbol size shouldn't exceed 26x26 or 28x28.

Maybe it could work as full baseplate design or on a 30x30 circle (similiar to what you did for Square).

You shouldn't do something like this
cinema4d-1

It's better to substract that thing from the blue shape. There's even more strange stuff going on in your Square icon, causing a jagged outline around the white circle.

really-need-an-api-key:
I haven't abandoned these two open PRs yet. Unfortunately I was a little busy lately, but I'll get to them soon.

Foggalong (changes requested):
As @palob said, symbol size needs reducing

Foggalong (approved):

PR #3551, opened 2017-01-15
fix pep8, rewrite some codes

bilelmoussaoui:
Replace paths from `theme/icons/sizes..` to `path.join("theme", "icons", "size")` This will allow users to build the icons in Windows and Unix systems. 
Also, I've removed unneeded code, duplicate code, cleaned variables that were not used or not needed =)
And added an input to ask the user to remove the old build if it's already exists
screenshot from 2017-01-15 21-03-33

`pylint gen.py` now gives 10/10 as a score =)
screenshot from 2017-01-15 21-02-27

### DIFF
--- a/gen.py
+++ b/gen.py
@@ -15,9 +15,9 @@ from json import load
 from os import listdir, makedirs, path, symlink
 from shutil import copy2, rmtree
 from subprocess import PIPE, Popen, call
-from gi import require_version
-require_version('Rsvg', '2.0')
 try:
+    from gi import require_version
+    require_version('Rsvg', '2.0')
     from gi.repository import Rsvg
     from cairo import ImageSurface, Context, FORMAT_ARGB32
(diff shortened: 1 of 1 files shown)


Foggalong:
Please separate out changes into different pull requests, I like most of what's being done here but there's a bit too much going on in one place.

bilelmoussaoui:
Well, the only thing that was changed is instead of using better variable naming and path.join instead of /. Other then that nothing was changed. I've just added a png2icns not installed exception (which is not really needed). Anyway, i will try to separate everything 👍

Foggalong:
@bil-elmoussaoui Fixing PEP8, rewriting sections, and improving functionality is too much going on for me - I'm very much a one pull, one topic kind of person :)

bilelmoussaoui:
Not a problem 👍 I will do that by this evening

PR #3560, opened 2017-01-21
Icons for Skifra

palob:
https://github.com/Skrifa/Skrifa

skrifa-128

skifra-numix


Foggalong (changes requested):
With the yellow as is there isn't enough contrast between the baseplate and the icon. Try either a darker yellow or an alt colour for the baseplate.

palob:
Here's a darker yellow.
skifra0

Foggalong (approved):

Foggalong:
I messed up there, really sorry about that - thought the change had already been pushed when it wasn't :P

PR #3562, opened 2017-01-23
Add symlink for plex-media-player

ccl2of4:
Just need a symlink for [plex-media-player](https://github.com/plexinc/plex-media-player). Unfortunately upstream does not provide their own `.desktop` file to be used across all distros, so I'm basing the icon name on the `.desktop` file from [this aur package](https://aur.archlinux.org/packages/plex-media-player/). Hopefully if any other distros choose to package this program they use the same convention.


bilelmoussaoui:
Thanks!

PR #3575, opened 2017-01-30
Only import needed libraries

bilelmoussaoui:
Import only needed libraries, and few verification. Next PR will use path.join in order to make the script works on other platforms other than that UNIX

### DIFF
--- a/gen.py
+++ b/gen.py
@@ -105,7 +105,7 @@ def convert_svg2png(infile, outfile, w, h):
 try:
     sizes = listdir("icons/" + theme)
 except FileNotFoundError:
-    exit("The theme {0} does not exists.Please reclone" 
+    exit("The theme {0} does not exists. Please reclone" 
         "the repository and try again.".format(theme))
 
 
--- a/gen.py
+++ b/gen.py
@@ -12,16 +12,15 @@ from argparse import ArgumentParser
 from io import BytesIO
 from json import load
 from os import listdir, makedirs, path, symlink
-from shutil import copy2, move, rmtree
+from shutil import copy2, rmtree
 from subprocess import PIPE, Popen, call
-
-from gi import require_version
-require_version('Rsvg', '2.0')
 try:
+    from gi import require_version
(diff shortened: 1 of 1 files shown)


gen.py
```diff
@@ -103,7 +102,11 @@ def convert_svg2png(infile, outfile, w, h):
 
 
 # Only certain icon sizes may be covered
-sizes = listdir("icons/" + theme)
+try:
+    sizes = listdir("icons/" + theme)
+except FileNotFoundError:
+    exit("The theme {0} does not exists.Please reclone"
```
Foggalong: There's a space missing between "exists." and "Please"

gen.py
```diff
@@ -12,16 +12,15 @@
...
 from os import listdir, makedirs, path, symlink
-from shutil import copy2, move, rmtree
+from shutil import copy2, rmtree
 from subprocess import PIPE, Popen, call
-
-from gi import require_version
-require_version('Rsvg', '2.0')
 try:
+    from gi import require_version
+    require_version('Rsvg', '2.0')
```
Foggalong, line 19: Just to clarify, are we only checking for Rsvg being installed within the Cairo `if` block because we're assuming if they have Inkscape installed they'll have Rsvg?

Foggalong (changes requested):

gen.py
bilelmoussaoui: What we do, is to check if the user have Cairo + Rsvg, if one or both of them are not found we fallback to use Inkscape instead (Rsvg is not needed by Inkscape)

gen.py
bilelmoussaoui: Done

gen.py
Foggalong: TIL Inkscape doesn't depend on Rsvg

Foggalong (approved):

gen.py
bilelmoussaoui: :P So it this ready to be merged?

PR #3579, opened 2017-02-04
Few fixes

bilelmoussaoui:
Use path.join instead of hardcoding the paths. This allow the script to be executable in other platforms without any issue 👍 
For now the script should be quite stable and works as expected. I was thinking of implementing the creation of the desktop files within the source code, the same as HiDPI support

### DIFF
--- a/gen.py
+++ b/gen.py
@@ -103,7 +103,7 @@ def convert_svg2png(infile, outfile, w, h):
 
 # Only certain icon sizes may be covered
 try:
-    sizes = listdir("icons/" + theme)
+    sizes = listdir(path.join("icons", theme))
 except FileNotFoundError:
     exit("The theme {0} does not exists. Please reclone" 
         "the repository and try again.".format(theme))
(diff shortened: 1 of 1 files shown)


Foggalong:
The one time I tried making this work on Windows all sort of things went wrong: file permissions were issued incorrectly, symlinks were botched, etc. Is there another situation where something other than `/` would be used in the file locations?

bilelmoussaoui:
All i can think of for now is the path separator :( It has been more than 7 years since i last used Windows as my main OS so i can't really test this correctly or debug it. If anyone willing to report issues about that, i will fix all of them :)

Foggalong:
Tbh, I'd vote just to not open it up to platforms that don't use `/`. It looks like a lot more trouble than it's worth for something that doesn't currently work, likely will never be used, and complicates the code.

bilelmoussaoui:
Anyway, my PR is more of a cleaning of the code than having a fully working windows version :)

Foggalong:
@bil-elmoussaoui There is some code cleaning, but most of it is switching over to `path.join` which is what I'm talking about

bilelmoussaoui:
@Foggalong Well, it's doesn't harm the code as it's works the same as before :) I can revert the changes to path.join if you wish =)

Foggalong:
@bil-elmoussaoui That's what I mean by what I said before: imo it does complicate the code for something that isn't needed. Reverting the change to `path.join` would be ideal.

Foggalong (changes requested):
@bil-elmoussaoui That's what I mean by what I said before: imo it does complicate the code for something that isn't needed. Reverting the change to `path.join` would be ideal.

bilelmoussaoui:
@Foggalong Instead of using path.join which i have removed in the latest commit, i have used .format, so the path is clear enough for us to see and debug in the future ;)

Foggalong:
@bil-elmoussaoui Now that's a change I can get behind! Will have a looksy asap :)

Foggalong (approved):

Foggalong:
Sorry for the delay, had a bit of a rough week with uni. Looks good to me though :+1:

bilelmoussaoui:
Not a problem :) i will get the travis PR by the next weekend !

Issue #3577, opened 2017-02-03
Icon Request - Writefull


leoheck:
https://writefullapp.com/

```
[Desktop Entry]
Version=3.0
Type=Application
Name=Writefull
Comment=Writefull helps you improve your writing
Exec=/usr/local/writefull/Writefull_3beta
Icon=/usr/local/writefull/Writefull.ico
MimeType=image/x-foo;
```

Original
![image](https://cloud.githubusercontent.com/assets/1277920/22595409/87dea2e8-ea0e-11e6-936f-edf53e01ffcd.png)

"My" idea from [here](https://writefullapp.com/tutorial.html). 
![writefull svg](https://cloud.githubusercontent.com/assets/1277920/22595312/1cc47960-ea0e-11e6-929b-3e729e4eef3e.png)

[My SVG source](https://www.dropbox.com/s/bwax3sx0s593eem/Writefull.svg?dl=0)

Foggalong:
@flexiondotorg If you want to open a pull request then we can start a proper review of your icon proposal :)

flexiondotorg:
Wha?

Foggalong:
@flexiondotorg Sorry didn't mean to summon you there! Must have misclicked, should have been @leoheck

leoheck:
Haha, ok I will do that @Foggalong

leoheck:
Done, it is now related to PR https://github.com/numixproject/numix-core/pull/3580

bilelmoussaoui:
ALready on master

PR #3581, opened 2017-02-04
Writefull

leoheck:
Only the Writefull icon


leoheck:
This resolves #3577

palob:
Can you please align all vertical lines to the pixel grid? Otherwise the icon will look blurry.

writefull

Did you get the symbol by vectorising the original icon (using potrace via Inkscape)?
Alas this doesn't yield useful results if there's straight lines or multiple colours involved.

Can you please use the current symbol as stencil and recreate it using simple paths (3-5 nodes)?

The symbol shadow should be coloured black at 10% opacity.

leoheck:
No, I just made a fast draft to discuss with you. I shoul apply all ideas you have. 
Let me know if I miss something.

Todos (improving the guidelines for the next time):
- Start from an existing icon
- Align vertical lines with pixel grid
- Create the symbol using paths
- Symbol shadow: single path, black, 10% opacity
- Baseplate: Vertical gradient +10 lighter (lightness) at the top

palob:
It's now composed of simple paths :+1: yet there are still misaligned vertical lines / vertical edges between differently coloured objects.

The shadow should be only one path. Can you please merge the paths it's composed of into a single one?

I think the baseplate (or the symbol) should be coloured differently. The icon lacks contrast right now.
What is more, the baseplate needs a vertical gradient and as such should be 10L lighter at the top.

The easiest way to get this right is to copy your symbol + shadow over to an exisiting icon.

gradient

You'll find these two handles when clicking on a baseplate of any of our icons (if not it's a bug). After clicking on the handles you can define the 2 stop colours of the colour gradient. The two colours should be equal except for the L (lightness) value which should be +10 for the upper stop.

leoheck:
Hm, good. I don't know if I understand pretty well but I will try to fix it. I also have started from an existing icon, I will look what I have messed.

palob:
By the way, we've got some (pretty hidden and incomplete) [design guidelines](https://github.com/numixproject/numix-wiki/wiki/Guidelines).

leoheck:
Please, put that link in the README file!

palob:
OK, now you've got the gradient but the colours are the same or not much better. The baseplate should be brighter (not brighter than our "white" icons) or much darker or you'd go for stronger colours on the symbol.

Alignment is actually worse than before. All horizontal/vertical lines need to be in line with the pixel grid (toggled with the ``#`` key in Inkscape).

By the way, there are two unneeded nodes on the shadow which you can delete.

palob:
Alignment still isn't spot-on. Same for symbol size and position.

alignment

leoheck:
I see, but this will change the symbol itself. I can improve it but I don't like to do that. And, if you zoom in the grid changes. Can you explain how this can improve the final icon? I don't know mutch about design to see why and how this is important.

palob:
If the border between 2 differently coloured areas isn't in line with the pixel grid (which really is a representation for the pixels on screen) it will come along blurry. 

That's because there is a line of pixels (highlighted in screenshot) with two colours assigned to it. A single pixel can display only one discreet colour so the colour shown will be interpolated.

(By the way the Inkscape grid only changes upon zooming because "looking from far away" there is no space for every single grid line. Instead you see one each 5px or so then.)

leoheck:
This is a good explanation, thanks for that. Did you saw my last updates?

palob:
It's properly aligned now :+1: . Can you move the symbol 1px down and get the symbol width to an even number. Otherwise it can't be centered horizonally.

I'm still not sold on the colours. Could you try something different than a white/greyish baseplate. Maybe the app's branding comes with other colours to draw inspiration from.

leoheck:
These colors are hard to make better, what do you think now?

palob:
The symbol definitely stands out better against the dark background. 
However now the baseplate lacks a gradient again.

There's fine-grained outlines between differently coloured areas in your system.
Two objects on top of each other musn't share the same edge:
overlap
Can you transform this into somthing like this?
overlap2
Morover if two bordering objects don't overlap at all the background will shine through, causing a fine-grained line. Can you make all objects overlap like this?
overlap3

Can you please recreate the shadow so it is aligned as well?

Last but not least, before we can merge them, the files need to be saved as "Optimised SVG" from Inkscape. (You might need to install ``python2-scour`` and ``python2-lxml`` for this to work). It is crucial that the CSS --> XML conversion option be NOT checked.

leoheck:
1. I couldn't understand the light version of the gradient to apply it by myself. I also started from another icon with a dark background. This icon doesn't have a gradient. I will try to improve it. Do you have any specifications for the gradient?
2. Sure, I will update overlapping shapes.
3. Recreate the shadow to keep it aligned.
4. And, optimize the SVG

palob:
In the colouring panel there is an option for linear gradients. Set the two stop colour handles to X0Y1 and X0Y47. The upper colour should be 10L (HSL notation) lighter.

Do you recall what's the icon w/o gradient?

leoheck:
Hm, I can't remember the icon and couldn't find it now.

palob:
Your progress isn't lost. The icon is up in git. Just clone your branch again and you get your latest commit.

Actually your icon is ready now.
All you need to do is to save the icons as *Optimised SVG* and a prophylactic symlink ``writefull`` (lowercase) won't hurt.

leoheck:
Actually, I don't know what have happened. Can you explain me? I already cloned it again. What I have to do to get the latest commit?

bilelmoussaoui:
@leoheck Symlinks shouldn't be added manually but using the data.json file

palob:
The shadows are misplaced. They should be 1px down + 1px right for Circle and 1px down only for Square.

Generally, the Square icon should have the symbol (shadow likewise) 1px higher.

There's something wrong with the baseplate bezel in your Circle icon.

This is yours
bezel-wrong

and this is how it's supposed to look like
bezel-right

Can you please copy symbol+shadow over to a fresh baseplate?

leoheck:
@bil-elmoussaoui, symlinks fixed 
@palob, I don't know if I could attend all of your questions, take a look, please. I have used the zsnes as reference.

palob:
Shadows are still misplaced, Circle baseplate bezel is still odd.

leoheck:
Does this mean that the zsnes is not so good?

leoheck:
@palob is it better now?

palob:
Actually there's nothing wrong with ``zsnes.svg``. Maybe you moved a bezel path above the baseplate accidently?

palob:
Shadows are still not where they should be.

palob:
We probably misunderstood. I was talking about the symbol shadow (which is still misplaced) while it seems you worked on the bezel thing  which is too light now.

palob:
Now that's it, thank you for your persistence :)

One thing left : the Square icon has the symbol slightly misplaced.

leoheck:
Good, and thank you for the instructions :)

leoheck:

palob:
The Square icon's symbol needs to go 0,5 px up, otherwise it's good to go. Thanks for your patience :)

palob (approved):
Good to go, thanks :+1:

Foggalong (approved):

PR #3582, opened 2017-02-04
Layout Editor (3D and Text Editor)

### LINKED ISSUE #3571: Icon Request - Layout Editor 3D and related
Starting the issue here as @palob requested, coming from https://github.com/numixproject/numix-icon-theme/issues/1180#issue-200999617

I have to create a new idea that can work better for smaller sizes. I also have to add a proposal for Layout Text Editor.

leoheck:
This resolves #3571


leoheck:

palob:
The icons do work for smaller sizes now but the symbol is spread over too much of the baseplate. Symbol size is about 31px X 33px which is definitely too much of a stretch. Moreover the **3D** and **TE** typeface objects need to be converted into paths.

You need to get the symbol smaller. Maybe it's better to drop the text-only design altogether and experiment with more creative metaphors for "text editor" and "3D".

leoheck:
Ok, what I have to do? Reduce the height, width or both? What is the maximum (tolerable) symbol size?

I like to have creative metaphors as you said, but I think it is nice to approach the author's intent. In this way, I have to think of an improvement.

palob:
26x26 (28x28) for non-spherical symbols, 30x30 for spherical ones.

Foggalong (changes requested):
As @palob said, symbol size needs reducing.

palob:
Smaller now but you need to tinker with the positioning. The letters aren't crisp enough and what is supposed to be an **L** is not recognisable as such.

Foggalong:
@palob You fancy having a look at this? Just realised it's coming up for a year since the last update

palob:
I can give this a shot. I envisage slightly 3Dfying the current ``layout`` icon.

palob:
No idea for the text editor one though.

palob:
Hmm, doesn't mix well with how we do shadows.

layout3d

I could try this in a "negative 3D", making the symbol look like engraved into the baseplate but then again this would be unlike any other icon.

Foggalong:
Yeah it's not a great original to work with :confused:

leoheck:
Haha, sure it is not original. It was a draft idea to put some icon on that app because the icon which came with de app sucks. I never thought it was going to take a year to move this thing forward. If you guys have a better idea, just use it. Actually, I like most when the icon recall the original idea (with colors or shape or symbol) of the app creators.

Foggalong:
@leoheck No worries, these things can be hard to get right. @palob It might be worth deviating from it entirely and just using some iconography related to the apps purpose instead.

maxtron95:
@Foggalong This issue has not been touched for 2 years. I have some ideas for Layout Editor 3D and Text Editor:

- Layout Editor 3D : Have a cube with the Letters L and E on the faces of the cube

- Layout Editor Text Editor : Have a notepad symbol with the Layout Editor Logo on it.

palob:
Yeah, the cube might do it!

leoheck:
🤡

Issue #2185, opened 2015-05-28
Icon for StarUML


delacruzbordajn:
A Numix-Circle Icon for StarUML is missing 

[Desktop Entry]
Name=StarUML
Type=Application
Categories=Development
Exec=/opt/staruml/staruml %U
Icon=staruml
MimeType=text/html;
Keywords=Design;Modeling;Development;UML;
![staruml_icon_by_funkmasterta-d2xh4wo](https://cloud.githubusercontent.com/assets/3194614/7852575/1c38abfc-04d6-11e5-8907-3803e9a142b7.png)
##

delacruzbordajn:
any notice?

Foggalong:
@delacruzbordajn Sorry for the big delay in replying! What sort of collaboration did you have in mind?

delacruzbordajn:
Any news about this?

Foggalong:
Issues take a while to fix these days, and there are ones older than yours still in line. Sorry

bil-elmoussaoui:
+1

Foggalong:
Please convert that +1 to a reaction on the original issue, then I will delete both your comment and this one. This prevents the issue tracker being clogged up with +1s

bil-elmoussaoui:
Done

nico-dela:
News?

palob:
Can you please open a pull request, @boosterdev?

PR #3584, opened 2017-02-04
Icon for StarUML #2185

walker-dev:


palob:
The symbol is a bit small. Could you try to expand it to 26x24 or even 28x26?

Currently there are fine-grained outlines between the different paths. You can get rid rid of them by having the paths overlap.

palob:
That's a bit bigger than what I had in mind but lets try with 30x28 in size located at X9 Y11 (Y12 for square). There's still some of those outlines, not all paths overlap and the place where all those "star beams" meet is not exactly X12 Y12.

I think the baseplate should be brighter and less saturated for better contrast. Moreover it needs a vertical gradient instead of a solid colour. The easiest way to get this gradient is to open an existing icon, deleting the symbol and pasting you symbol in. You can edit the baseplate gradient by clicking on the baseplate. You'll see two handles:

gradient

Clicking those lets you edit the colours. Both colours should be the same except for the L value in HSL notation where the upper colour should be +10L lighter compared to the bottom colour.

PR #3589, opened 2017-02-10
Icon for Encryptr

gregorydk:
New icons for Encryptr


palob:
Symbol should be 1px further down but it's a bit big with 30px height anyway. Can you try to get a smaller size working?

gregorydk:
Really?
1px down? It's at the same height as the semaphor icon. If I lower it I'll have to redo semaphor too.
Too big? It's the same size as the semaphor icon. If I change it I'll have to redo semaphor too.

palob:
Oops, so the semaphor height escaped me. I don't think you'd need to redo the semaphor icons if you shrinked this one.
The Encryptr icons looks much more heavy (it's actually bigger as in area).

Definitely 1px down. It might be centered numerically but due to being asymmetrical it looks off centre right now.

gregorydk:
I will change the Encryptr icon but I will also redo semaphor with the same changes. They're part of a set build with triangles of the same size.

palob:
Yep, that's it!

palob (approved):
It's good to go :+1:

Foggalong (changes requested):
The symbol in the circle icon needs moving down 1px to counteract the triangle illusion at the bottom. The square icon is fine though :+1:

PR #3593, opened 2017-02-15
Resize icon for Semaphor

gregorydk:
Resize symbol to be consistent with Encryptr.

See https://github.com/numixproject/numix-core/pull/3589#issuecomment-279234925


palob:
Thanks!

Issue #2604, opened 2015-10-15
Icon for slack


takayukioda:
A numix-circle icon for slack is missing
The icon line says, `Icon=slack` and original icon is attached.

![slack](https://cloud.githubusercontent.com/assets/1546410/10502123/9eaa45ee-7323-11e5-8915-8db39cfdd3d5.png)

Thanks!
##

TamasBarta:
It isn't missing, it's doesn't look like Slack. It uses the "hash mark" logo of Slack. (Which is great for the panel, but not appropriate as a launcher icon, I believe)
I'd also love to see this icon's numixified version as Slack's launcher icon.

ediTLJ:
Any hopes of seeing this done?

bilelmoussaoui:
Fixed

PR #3608, opened 2017-03-03
Slack

leoheck:
Improving circle slack icon following issue https://github.com/numixproject/numix-core/issues/2604 and adding square version.


palob:
+ Your icons use clipping which is a no-no since it doesn't work with Qt.
Can you please transform the clipped paths into regular paths using the intersection tool?

+ The coloured objects need to overlap in order to avoid the outlines.

+ The **S** should be better aligned to the grid

+ Please save the files as *Optimised SVG*

palob:
There's a greyish baseplate underneath the new stuff causing an outline around the baseplate.
Can you please remove it?
There's something wrong with the bezel again.

Can you please the **S** object such that its XxY dimensions are integer px values (same for X and Y positions)?

All in all the **S** looks a bit wobbly. Can you simplify it?

leoheck:
@palob is it better now?

palob:
Baseplate is fine now :+1: 
Can you resize the **S** to 12x16 and place it at X18 Y16?

The **S** is still a bit wobbly. How did you create it?

leoheck:
I started from an svg icon which I found on google.

leoheck:

palob:
(There's some minor issue with the "rectangular" shapes. But I can sort out that later by myself.)

There's one edge without overlap left in the Square icon, causing an outline. Can you please fix that?

leoheck:
I would like to know what is the issue with rectangular shapes. Can you explain?

palob:
Ideally, the rotated white square would exactly be 28x28 in size and be located at (10,10). Angle of rotation is something like 18.79°.

Actually, in the Square icon everything is bigger. Is this on purpose? White Square should be 30x30 at 9,9 here.

leoheck:
I started both icons from a squared version. However, the round corners of the original were different of numix-squared, so, in order to fit it in our shape, I had to increase the size of the original icon. Circle and square versions should be equal?

leoheck:
I would like to fix it, but I don't know how to manage the corners properly. If you have any advice, I would like to know. I made the white square 30x30@9x9 but I have to change all other shapes to align the white square. I can select and move but this will mess with the corners.

palob:
To get the corners right you just need to do an intersection of a rotated square and the baseplate.

leoheck:
Hm, this is the same thing I did, but ... I didn't realize that it could be used to solve other issues. Haha, thanks.

leoheck:
This was easier than I expected. I made it from the circle, let me know if it is good.

leoheck:
Oh, you was talking about the circle icon here, right?
> Ideally, the rotated white square would exactly be 28x28 in size and be located at (10,10). Angle of rotation is something like 18.79°.

So, I have to fix them.

palob:
Square size and position is OK now but in the Square icon there are lots of unneeded nodes now.

palob:
There's still some left, for instance here:
nodes

leoheck:
Yes, I don't know how to handle them near corners yet.

palob (approved):
Better now and ready to go!

Foggalong (changes requested):
In the square icons the dropshadow goes immediately below the symbol, not to the right of it as in circle

palob:
@Foggalong That's actually how it's supposed to be, isn't it?

Foggalong:
@palob It is now yes :+1:

Foggalong (approved):

leoheck:
When slack will appear in Ubuntu repositories? I still have the same icon. Using ubuntu 17.04

palob:
By the time we'll have a new [release](https://github.com/numixproject/numix-icon-theme-circle/releases).

PR #3616, opened 2017-03-11
Added new symlink for Netflix Chromium App

heyhippari:
I added Netflix as a Chromium app on Fedora 25 and noticed the icon wasn't themed because it used a different name, so I added that to the symlinks of the existing Netflix icon.

### DIFF
--- a/data.json
+++ b/data.json
@@ -8706,6 +8706,7 @@
             "symlinks": [
                 "chrome-ahdabofbdlkjeniniajfcdlhanijbakd-Default",
                 "chrome-deceagebecbceejblnlcjooeohmmeldh-Default",
+                "chrome-cgbnengjfcijdonppjmgiodcdincmebo-Default"
                 "netflix",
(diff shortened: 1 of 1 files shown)


palob:
You forgot a comma :)

heyhippari:
Oops, thanks for pointing it out :p

Seems like Chrome changes the hash every time you add a website, though, so merging this in might be a bit useless...

bilelmoussaoui:
@MrTimscampi Maybe pointing that out to Google ? So they can use a hash based on the url or something less variable !
Thanks

palob:
@MrTimscampi So is this symlink of any help?

heyhippari:
@palob Seems after testing that none of the Chrome-*-Default for websites added through "Add as an app" symlinks are useful... It generates a new hash every time you add a site, so nobody will have the same desktop entry.

I think this can safely be closed.

Issue #3150, opened 2016-05-15
Icon for Duet (Steam)


matemeszaros:
Game: Duet
AppID: 292600
Icon: ![duet-icon](https://cloud.githubusercontent.com/assets/15245298/15275538/d439c0f6-1ace-11e6-893f-926d371d1b6b.png)
Notes: Game about maneuvering. No default in steam, this one is copied from a google images search - should not be stripy.

mrmeszaros:
Created a simple reference SVG:

![duet](https://cloud.githubusercontent.com/assets/15245298/21118138/ddb8e44c-c0bc-11e6-9118-250459fd757e.png)

And the code:

``` svg
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg xmlns="http://www.w3.org/2000/svg" version="1.1"
     viewBox="0 0 48 48" fill="none">
  <defs>
    <linearGradient id="blue-shadow" stop-color="#33BBFF"
                    x1="50%" y1="15%"
                    x2="50%" y2="75%">
      <stop offset="0" stop-opacity="0"/>
      <stop offset="1" stop-opacity="0.5"/>
    </linearGradient>
    <linearGradient id="red-shadow" stop-color="#EE1111"
                    x1="50%" y1="25%"
                    x2="50%" y2="85%">
      <stop offset="0" stop-opacity="0.5"/>
      <stop offset="1" stop-opacity="0"/>
    </linearGradient>
  </defs>
  <g transform="rotate(-30 24 24)">
    <!-- Background -->
    <circle cx="24" cy="24" r="23" fill="#111111"/>
    <!-- Red shadow -->
    <path d="M24,8 A16,16 0 0,0 24,40 A11,11 0 1,1 24,18" fill="url(#red-shadow)"/>
    <!-- Blue shadow -->
    <path d="M24,40 A16,16 0 0,0 24,8 A11,11 0 1,1 24,30" fill="url(#blue-shadow)"/>
    <!-- Red circle -->
    <circle cx="24" cy="13" r="5" fill="#EE1111"/>
    <!-- Blue circle -->
    <circle cx="24" cy="35" r="5" fill="#33BBFF"/>
  </g>
</svg>
```

mrmeszaros:
Fixed.

PR #3618, opened 2017-03-15
Icon for Duet (Steam)

mrmeszaros:
Added a steam icon for the game Duet (issue #3150).


bilelmoussaoui:
Thanks for your PR. You will need to include a square icon too, and also modify the data.json file to include your icon. 
The themes are generated later by @Foggalong  using the `gen.py` script. 
If you need some help just ping me.
@palob will take care of the design review

palob:
Hi, thanks for your PR.
Some comments on your icon:

+ The baseplate needs a vertical colour gradient (10L lighter at the top). Please take a look at existing icons for reference. You can copy your symbol+shadow stuff over to an existing icon and alter the gradient stop colours.

+ Can you please shrink your symbol down to 30x30 pixels (32x32 currently) and align the circles to the pixel grid afterwards (integer px size and integer px coordinates)?

+ On the Square icon the symbol will be one px higher and the symbol shadow is only 1px down from the symbol (no horizontal shift)

mrmeszaros:
Hi, thanks for the comments.
I would like to have some questions though:
 - By 10L lighter you mean 10% lighter (in HSL space)? That seems to me a little too much - considering that some other steam icons I checked have about 4-5% lighter only.

 - I will gladly shrink the symbol down to 30x30 px (I had a slight feeling that it might be a little to heavy). I am using now a -30° rotation, should the coordinates in the sources be on the grid (before the rotation), or the resulting ones (after the rotation)?

- I will start making the Square icon based on the template.

palob:
+ 10L absolute in an 8-bit (0-255) HSL scale.

+ Not sure what you mean since a circle's coords aren't changed through rotation. It's important that the circle be aligned to the pixel grid (``#`` in Inkscape) in the finished icon. Otherwise it will be blurry.

mrmeszaros:
@palob does the Square icon need the background gradient?

Also, I force pushed the requested amendments, I hope it is okay.

palob:
The symbol size is 30x30 now and the circles are 8x8 pixels big but there still aren't placed at integer coordinates. This way they aren't aligned with the pixel grid as a result of which they look blurry.

circles

I don't think the colour contrast is ideal. Could you try a lighter red?

mrmeszaros:
Thanks for the comment - you are right, forgot the snapping.

I can snap the circles to the grid. But I am now struggling with the shadows with the gradient.
The shadows so far were bounded by circle (elliptic) arcs, but Inkscape has turned them into Bezier-curves (instead of changing the radius).

Also, I made the red a bit lighter and more yellowish, what do you think?
image

mrmeszaros:
I ma not able to make the red/blue shadows circular.
If the oval shape does not look good, the balls can be vertical  too (this way things line up perfectly):
image image
What do you think?
(Oh just noticed I did not copy the symbol shadow - please try to ignore it)

palob:
In case you meant to push changes - you didn't.

I think I prefer the new red. I don't understand what you mean by "oval" and not getting the shadows circular. (By the way when I talked about symbol sizes this is always excluding the shadow) 

Regarding your new tilted design (the right one from above), I like that as well.

palob:
Everything nicely aligned now.

Can you please merge all shadow elements into one path and remove all comments from the SVGs (they cause issues with QtSVG)?

mrmeszaros:
Some parts of the shadow have linear gradients (the ones of to the red/blue gradients).
Merging them into one path would mean that I have to remove the gradients (the two gradients have different directions).

Tried it, did not look so well:
image

Other than that - should I convert other non-`path` elements (`circle`-s) to paths?

palob:
You can add a third stop to the gradients. This way you can keep the colour constant for the circle area and merge the shadows into only two paths (circle-swirl pair as one path).

There shouldn't be any overlapping shadow objects since their opacities add up in Qt.

palob:
Actually you did what I had in mind but alas it looks like this in Nomacs 3.6.1 (a Qt image viewer) with Qt 5.8.0:

icon-in-nomacs
(same for Circle)


I don't exactly recall what's the reason for this particular rendering bug but in your SVG source the linear gradients are defined with relative coordinates, inherited colours and XML attributes.

How did you save the files? Did you use any SVG cleaner?
You'll probably have to redo the gradients and save the icons as **Optimised SVG** from Inkscape while being sure the **CSS to XML** attributes conversion option be not checked.

palob (approved):
Rendering fine now, thanks

mrmeszaros:
- I think I have fixed the linear-gradient issue for Nomacs.
- I usually edit these files manually from sublime-text, and  preview it with Firefox (and after your comment with Nomacs).
- Did not use any SVG-cleaner.
- If needed I will transform the XML attributes to CSS (probably with [Scour](http://codedread.com/scour/), or [XSLT](https://gist.github.com/matemeszaros/500a151d5a712165d1ccff46b227c4c7))
- I just found that the game is also available for [Android](https://play.google.com/store/apps/details?id=com.kumobius.android.duet).
- Is it possible to add this icon (if finished) to the Android icon pack (e.g.: with some `data.json` config)?

bilelmoussaoui:
@matemeszaros For now, in order to get the Android support to work correctly, we will have to update the android repository and get all the icons from there and put them on this repository, add those icon names to the data.json and create a new release. Hope this will happen before this summer :+1:

palob:
@matemeszaros You don't need to check with Firefox (or any other browser). Just make sure it looks fine with QtSVG (for instance Nomacs, gwenview) and rsvg (GTK+, for instance Eye of GNOME, Eye of MATE, GThumb).

PR #3634, opened 2017-04-05
move android icons to numix-core

bilelmoussaoui:
This PR contains the icons from the Numix Android Circle icon pack. Some of the icons might need some modifications or fixes.

Before this can be merged: 
- [ ] Be sure that none of the icons are a duplicate of existing icons
- [ ] Update/clean all the icons
- [ ] Update the data.json file 
- [ ] See if none of those icons can fix existing issues on the numix-core repo (symlinks maybe?)

Maybe? 
- [ ] Create square version of each icon 

@palob @Foggalong What do you guys think about that?


Foggalong:
We do already have Square, Hexagon, and Fold icons for Android. I can bring those across :)

bilelmoussaoui:
Can you do that on the same branch? Ping me if you need some help :)

palob:
Will they go into the same directory in core? They're designed based on different guidelines (or not up to snuff) mostly.

Foggalong:
@palob They go into the same directory because in theory the design guidelines are exactly the same. Anything not up to snuff should be marked for redesign though (it might be easier to just redesign those that need it in this pull request?)

bilelmoussaoui:
@Foggalong do we need the other icon packs or we should support only the the circle/square ?

Foggalong:
@bil-elmoussaoui I'll tell you want, I'll leave all of the themes except Circle & Square out of this pull request and then readdress it all later.

We don't need them, and the themes they produce won't be supported, but I think I'm gonna include the icons here at some point just for consistency. I might also do the same with uTouch and Shine also.

bilelmoussaoui:
Yeah I will take care of the square version then !

Foggalong:
I'm assigning myself so I remember I actually have something to do on this pr :P

bilelmoussaoui:
@Foggalong Especially after this PR https://github.com/numixproject/com.numix.icons_circle/pull/50 :+1:

Foggalong (changes requested):
Looking through the Square Android repo I think it'll honestly be quicker to just fix all the Circle icons and then make the Square version from those, rather than adding and fixing the Square version separately. It's a mess of out of date designs, weird quirks, and horrible compression :cry: 

On the positive I've started the process of adding the Android icons to the `data.json`, including all of the ones I could find which were just Android version of existing Linux icons. If anyone else fancies getting involved with this then feel free by either pushing to the Android branch directly of filing a pull request against it.

bilelmoussaoui:
@Foggalong That would be good too :) For now, i'm quite busy but i will do some work on that whenever i have some free time

W-Floyd:
Making no promises, but I might give this a poke soonish™, if that's alright. It could take a while, depending on how much cleaning up the old icons will need. Just need to parse a list of icons added here, and I'll make a start.

Edit: Hmm... I'm not entirely clear on which icons need what done to them, so I'll start on what I think are the right ones, and hopefully someone can clue me in later.

Edit 2: I'm using #3895 as a base for what I'll be doing, let me know if that's an issue.

bilelmoussaoui:
@Foggalong I think that moving the whole icons was a very bad idea. It will be really a hard task to review and merge. I propose to close this PR and #3895  and add few icons by PR, this will allow an easy review and we can move on this faster to release a new version of Android icon packs soon

Foggalong:
Agreed :+1:

Issue #2659, opened 2015-11-12
Icon for The Stanley Parable [$5]


bil-elmoussaoui:
The stanley Parable is a steam game!
The icon line : `steam_icon_221910`
The default icon 
![steam_icon_221910](https://cloud.githubusercontent.com/assets/7660997/11128735/f33c6560-897b-11e5-82ec-985eb78b70cb.png)
##

bil-elmoussaoui:
I added a bounty on this as i love this game and it's one of the missing icons in my desktop!

mrmeszaros:
Hey @bil-elmoussaoui,
I made a draft based on the default icon:
![stanley_parable svg](https://cloud.githubusercontent.com/assets/15245298/24811721/8124152e-1bc7-11e7-97ce-3b521652e8fb.png)
It could also be based on the iconic terminal:
![](http://orig00.deviantart.net/9502/f/2013/293/f/2/the_stanley_parable___icon_by_blagoicons-d6r6i19.png)
What do you think?

bilelmoussaoui:
The icon looks great! I will let @palob @Foggalong give you a better report :+1:

palob:
@matemeszaros Looks a bit small on the baseplate but I'm looking forward to your PR.

mrmeszaros:
@palob Here is a scaled up version (scaling factor: 1.2, ended up with about 29px symbol hight):
![stanley_parable x120](https://cloud.githubusercontent.com/assets/15245298/24829523/a4d4bb34-1c73-11e7-8729-e1a51aa8f874.png)
Brace yourself, the PR is coming :wink:

bilelmoussaoui:
@mrmeszaros You can claim the bounty :1st_place_medal:

mrmeszaros:
@bil-elmoussaoui Although the bounty is listed on https://www.bountysource.com/teams/numix - the link is broken :unamused:

bilelmoussaoui:
@mrmeszaros I have the same issue too :( even if i'm the one who put the bounty on this issue

PR #3636, opened 2017-04-08
Icon for The Stanley Parable

mrmeszaros:
A rough icon base for The Stanley Parable steam game (issue #2659).


Foggalong (changes requested):
Design as it currently is looks distinctly un-Numixy. I've not had chance to play around it, but guessing it has something to do with the hand

palob:
You've got an outline around the different parts. Could you please either try a thicker outline or no outline at all (i.e. no free space between the different objects)?
I still think it looks small and detailed. Maybe it helps to go without the lower tie part and without the sleeve.

The tie could do with a bit less saturation / more brightness.

PR #3638, opened 2017-04-09
Icon for Icemon

palob:
https://github.com/icecc/icemon

hi128-app-icemon

icemon-numix


Foggalong (changes requested):
The contrast between the yellow circle and cream background isn't really high enough. Fixing this might involve finding a different background colour because I think changing the yellow to work might necessitate it be brown.

palob:
One of these?
icemon

Foggalong:
0 works well :+1:

Foggalong (approved):

PR #3644, opened 2017-04-10
Add square icon for Avocode

heyhippari:
Part of numixproject/numix-core#3481

Edit: I saved it as Optimized SVG in Inkscape, but it seems that it left some. Should I fix it ?


palob:
Thanks for contrubuting again :+1: 
The icon is all fine however I am not sure we can merge verified commits due to a launchpad bug.

@Foggalong @wa4557 @bil-elmoussaoui Didn't Numix base run into PPA issues because of verified commits?

heyhippari:
Oh, if that's an issue, then I can refork and commit again, signing them is just a habit that I took :)

palob:
It's probably best to do this prophylactically, fixing it was a real pain. Thanks again.

andia89:
Yes no commits that are signed, launchpad can't import signed commits from git. 

On a sidenote: I hope to be around more often soon, work (mainly) is keeping me a busy atm

PR #3664, opened 2017-04-17
Icon for PhotoFiltre-LX

palob:
https://github.com/PhotoFiltre-LX/photofiltrelx

logo
filtrelx


Foggalong (changes requested):
@palob Is there a way you can get more contrast with the blue and green? As is they kina look like one piece

palob:
Pushed this
photofiltrelx

Foggalong (approved):

PR #3668, opened 2017-04-18
Symlink for plexmediaplayer

heyhippari:
This is the replacement for Plex Home Theater, so the icon is the same.


bilelmoussaoui:
Thanks !

PR #3677, opened 2017-04-19
Icon for Kakoune

irx:
kakoune
Icon for Kakoune, a vi-like text editor [http://kakoune.org](http://kakoune.org)

I'm wondering whether to make the symbol smaller...


bilelmoussaoui:
Before discussing the icon design, can you please add a square version of the icon? and also modify the data.json file to include the new icon name! Thanks

palob:
Thanks for contributing.

Yeah, the symbol is too big. Please check 26x26 or 28x28 at max.

Can you please make the slightly slanted edges straight in order to achieve alignment with the pixel grid?

kaoune

As for the Square icon, the only difference is that the shadow is only 1px down relative to symbol and both symbol and shadow go 1px up compared to Circle.

Actually I prefer the colours of the original icon. I think they provide better contrast and look less garish.

palob (changes requested):
Thanks for the Square icon. In addition to the size and colour suggestions all shadow objects in the Square icon need to be moved 1px to the left. For reference please see existing Square icons.

palob:
Size is fine now :+1:  however the **K** has to be realigned with the pixel grid. Taken as a whole the colours look darkish and really need more contrast. Currently the icon is not reconisable at smaller sizes.

irx:
Sorry, I've made this square icon before reading your post, so that's why it's this way :/
I will however correct this as well as 'straighten' that K, though the original logo is deliberately slanted, but as you noted it may cause problems with readability, so I'll make it more regular K

palob:
Symbol colours are better now but the baseplate still clashes I think.
Can you try a darker, less saturated colour like on the webpage or a significantly brighter colour?

You can remove the upper left element of this shadow path and merge it with the other oter shadow path.
kaoune

irx:
kak1
kak2
I'm having hard time deciding which colour is better: the one taken from the original logo or that olive from generic palette. :stuck_out_tongue: 

...but I guess I'll go with the former.


Then there's a thing that's bothering me - those backplates happen to have some sort of gradient which
visually seems to be flat colour. Should I replace colour values in this gradient xml definition or just make a simple fill?

palob:
Our baseplates do indeed use a vertical gradient (10L lighter on top).
You can change the stop colours by clicking on the baseplate of an exisitng icon. See discussion and screenshot in #3683.

palob:
Gradients are in place :+1: It appears the Square icon has been optimised too much. The SVG source is squeezed into 1 line. This should be avoided because it renders the source code unreadable.
(Many Square icons still are one-liners)

I'm still not entirely sold on the colours. Can you check out if you can find a colour combination which makes the **K** and the rhomb stand out better? 
In the original icon this is done by the help of gradients.

@Foggalong Any ideas for colous?

Foggalong:
@palob I actually quite like the colours pushed in the most recent commit. It'd be nice if there was a little bit more contrast between the blue part of the K and the segment of the rhombus below it, but I played with a few colours and couldn't find anything I liked more than the ones @Ewaryst used.

palob:
Yeah, I had fiddled about a bit as well and couldn't find anything that much better.

So it's gut to go except for the one-line Squre icon.
@Ewaryst Can you please fix it?

irx:
Is indentation preferred or is it better the way it is now?

palob (approved):
It's OK now, thanks for the rad icon!

irx:
Pleasure to contribute :smile:

PR #3678, opened 2017-04-19
Square icon for Kodi

heyhippari:
Part of numixproject/numix-core#3481

I resized the symbol to fit into the circular symbol guide, as I thought it looked better than fit to the square symbol guide (It looked to small).


palob (changes requested):
The vertical lines aren't in line with the grid. As for the bigger size, please transform the symbol to a full 30x30 instead of 29.something px.

palob:
kodi


Please remove the comment in [line 2](https://github.com/MrTimscampi/numix-core/blob/acfb8ae29ee291e599f6613bfe3fe8ed7de0443e/icons/square/48/kodi.svg?short_path=d80e0a9#L2), it causes problems with Qt.

heyhippari:
For consistency's sake, I reverted the icon size (The geometry would have to change a bit for the resizing, which isn't really optimal).

I also added the 10L gradient on the baseplate since, apparently, it's missing from the square template and I hadn't noticed (I'll make a pull request to fix all my previous square icons after this).

palob:
Thanks for the fixing.

palob (approved):

PR #3683, opened 2017-04-20
[Square] Enpass

bilelmoussaoui:
From #3481 
enpass


palob:
The baseplate has to have a vertical gradient instead of a solid colour (10L lighter at the top in HSL notation). You'll find the 2 handles to adjust the stop colours of the gradient when clicking on the baseplate of an existing icon (usually left to the icon, same for Square or Circle).

handles

You can either modify an existing gradient or add a linear gradient manually and drag the stops to where they belong.

Probably all of your Square conversions lack a gradient but I can't check because I'll be off for a few hours now.

bilelmoussaoui:
Alright, I will take care of that:+1:  Thanks !

bilelmoussaoui:
Done for the linear gradient

palob (changes requested):
You've got a gradient now but it is the wrong way around :) The baseplate should be lighter on top.

bilelmoussaoui:
@palob I've made an attempt to fix that :) Please let me know if i should modify something else, if everything is okay i will do the same for the rest of my PR's

palob (approved):

Foggalong (approved):

PR #3693, opened 2017-04-22
Icon for heXon

palob:
https://github.com/LucKeyProductions/heXon

hexon-ori

hexon-numix

### DIFF
--- a/data.json
+++ b/data.json
@@ -6274,6 +6274,11 @@
             ]
         }
     },
+    "hexon": {
+        "linux": {
+            "root": "hexon"
+        }
+    },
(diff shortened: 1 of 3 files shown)


heyhippari:
I just noticed the square one is still named hexon-sq.svg in the repo. Is that normal ? Shouldn't it be hexon.svg ?

The circle version also isn't saved as Optimized SVG.

palob:
Thanks for reporting, will fix right away :)

heyhippari:
You're welcome, I noticed the name and the higher size while updating my fork, so I figured I might as well report it ;)

PR #3695, opened 2017-04-23
Icon for Office Runner

palob:
https://git.gnome.org//browse/office-runner/

office-runner

office-numix


Foggalong (changes requested):
Square down 1px I think

palob:
Absolutely. 1px down it is :)

PR #3720, opened 2017-05-17
New Zotero circle icon - closer to official logo

R-Broadley:
Changed background from Burgundy to Grey and Z from white to red.
Z also made bolder.

Official Zotero logo can be found here: https://www.zotero.org/support/_media/logo/zotero_512x512x32.png

| --- | :-- | :-- |


Foggalong (changes requested):
You can change the colours to match the original icon, but please retain the shadow. That's part of the Numix guidelines.

Also, would you mind adding a square icon in this pr too?

R-Broadley:
Square icon and shadows are no problem, will add them.
Where are the Numix guidelines, can't seem to find them?

Foggalong:
@R-Broadley They're [officially here](https://github.com/numixproject/numix-wiki/wiki/Guidelines), but are also very out of date. I keep meaning to write new ones but then other stuff comes up. In the mean time I'd recommend just looking at other Numix icons and taking the key features of them (shadows, gradients, etc) into yours. We're also more than happy to help address any problems that come up :)

palob:
Hi, what needs fixing above all is that vertical and horizontal lines need to be aligned with the pixel grid (``#`` key to toggle it in Inkscape). Otherwise the icon will look blurry.

zotero

R-Broadley:
I think I've addressed all the comments, shadow re-added, aligned to pixel grid, square version updated.

bilelmoussaoui:
@R-Broadley The shadow of the square icon must be moved to 1px on the left

R-Broadley:
Shadows should be identical on both (the Z translated (1, 1) with black color, 0.1 opacity).
What are you viewing them with?

palob (changes requested):
Symbol is still misaligned

R-Broadley:
@palob some specifics would be nice, misaligned in which direction and which icon, circle, square or both.

bilelmoussaoui:
@R-Broadley I can't visualize the icon on Inkscape :/ 
screenshot from 2017-05-18 22 21 05

palob:
Relative to the pixel grid like for instance in the current icon.
zotero

Any pixel which is filled up with more than one single colour will be shown with an interpolated colour.
If a boundary between a red rectangle and a yellow rectangle isn't congruent with the pixel grid a orange line will be shown between the two on screen. This is where the blurriness originates from.

R-Broadley:
@bil-elmoussaoui it appears inkscape does not have support for svg's lineto commands (https://www.w3.org/TR/SVG/paths.html#PathDataLinetoCommands)
The file renders everywhere else just fine.

R-Broadley:
@palob I'm not following, the most recent version should line up with the pixel grid (e.g. the top of the z is 5px wide from 14, 14 to 32, 14). Not that it makes any difference when it is inevitably scaled.

palob:
No, the 48px preview is still blurry and just like @bil-elmoussaoui in Inkscape I can't see anything of the symbol at all.

R-Broadley:
Should all be fixed now, I don't normally use inkscape but have checked the latest version of the icons and they now display correctly as far as I can see.

palob (changes requested):
You implemented the symbol (as well as the shadow) as stroke. Can you please turn it into a path?

Foggalong:
@R-Broadley Can you give us permissions to push to the branch these changes were made on so we can finish the work and get it merged in?

PR #3726, opened 2017-05-20
Icon for Reigns (Steam)

mrmeszaros:
| --- | :--: | :--: |
|  original reigns | reigns circle | reigns square |


mrmeszaros:
The symbols were clearly too big, so i shrunk them a bit:

| Circle icon | Square icon |
| :--: | :--: |
| reigns circle | reigns square |

palob (changes requested):
It's a cool design but I think it could still be samller. You can get the symbol height down to 28px by shortening the merlon things by 1px, slimming down the "castellation base band" to 4px height and shortening the dower base by 1px.
There's some strange inner paths in the 2 yellow castellation paths you can get rid of by performing a path union on each path itself.

palob:
Size is fixed but there's still the inner paths (same for the shadow).
innerpath

palob (approved):

PR #3752, opened 2017-06-05
Bunch of square icons

andia89:
Yeah with the templates it is still tedious but doable in a finite amount of time...
I moved the symbols up 1px in a few icons where it seemed reasonable to me


bilelmoussaoui:
The icons should be saved as optimized SVG

palob:
Yeah but without the CSS to XML conversion option.

andia89:
Ran svgcleaner instead

palob:
The latest git version with the option ``--join-style-attributes=all``?

Otherwise the icons will suffer from the gradient bug.

palob:
Icons as such are fine. I wonder if all icons would look better with the symbol 1px up.

andia89:
Do you think?
screenshot from 2017-06-11 13-23-35

For me some of them are already too far up...
But we should figure that out, before I'll add more.

palob:
For the Square icons I think the bottom bezel isn't part of the baseplate. It's basically the side if you look at it 3Dish.

Foggalong:
@wa4557 This seems to have ran into a conflict because some icons were covered by another commit which is missed. My bad :disappointed: :

Foggalong (changes requested):
This seems to have ran into a conflict because some icons were covered by another commit which is missed. My bad :disappointed:

andia89:
Closing since @W-Floyd did this amazing job with the square icons

W-Floyd:
lol Now I can stop going from reverse alphabetical order :3

Issue #3482, opened 2016-11-07
Optimisation & Design Updates for Square


Foggalong:
Of the icons already covered in Square, not all of them have been properly checked over. This includes:
1. Have the latest Numix design (the one used by Circle)
2. Be void of clipping, strokes, and other elements which cause problems
3. Be saved as an Optimised SVG

This will be done as new icons are covered for #3481 but the existing ones need going over too, which is what the list below covers.

* [x] 0ad
* [x] 2048
* [x] 3Depict
* [x] 4kvideodownloader
* [x] 4Pane
* [x] 7zip
* [x] 8-ball-pool
* [x] 9AD6_Origin.0
* [x] aarddict
* [x] abgx360gui
* [x] abiword
* [x] abrowser
* [x] accerciser
* [x] accessibility-directory
* [x] accessories-camera
* [x] accessories-dictionary
* [x] accessories-document-viewer
* [x] accessories-image-viewer
* [x] accessories-media-converter
* [x] accessories-painting
* [x] accessories-screenshot
* [x] accessories-system-cleaner
* [x] access
* [x] aceofspades
* [x] acestreamplayer
* [x] acetoneiso
* [x] acroread
* [x] actionaz
* [x] activity-journal
* [x] addressbook
* [x] administration
* [x] AdobeAfterEffect
* [x] AdobeAIR
* [x] AdobeAudition
* [x] AdobeBridge
* [x] AdobeDreamweaver
* [x] AdobeEncore
* [x] AdobeFirework
* [x] AdobeFlashBuilder
* [x] Adobeillustrator
* [x] AdobeIndesign
* [x] AdobeLightroom
* [x] AdobePhotoshop
* [x] AdobePrelude
* [x] AdobePremierePro
* [x] AdobeSpeedgrade
* [x] AdobeUpdate
* [x] AdobeWidgetBrowser
* [x] aegisub
* [x] agave
* [x] agenda
* [x] ailurus
* [x] AirPortUtility
* [x] akonadi
* [x] albert
* [x] alchemy
* [x] alevt
* [x] alexandra
* [x] alienarena
* [x] alleyoop
* [x] almanah
* [x] AlphaPlot
* [x] altyo
* [x] am2r
* [x] amarok
* [x] amazonmp3
* [x] amazon-store
* [x] amd
* [x] amnesia-dark-decent
* [x] amsn
* [x] amsynth
* [x] amule
* [x] anatine
* [x] android-arc
* [x] android-file-transfer
[...]

PR #3757, opened 2017-06-13
Found and optimised any files with "Created with Inkscape"

W-Floyd:
Don't know if I did the right thing, but I went through and optimized any files that contained "Created with Inkscape".


palob:
Which commands did you run?

andia89:
I would prefer something that deletes only the appropriate line, honestly.

`find . -type f -iname *.svg -exec sed -i '#<!-- Created with Inkscape (http://www.inkscape.org/) -->#d' {} +`

should also do the trick (the string needs to be escaped properly probably)

W-Floyd:
Hmm. I believe it was scour, and sed to remove the Inkscape bit. However,
I've been seeing some different instructions on what settings to use, so I
may redo this. I guess the question is if it's even necessary, and welcome.
If it's not, then feel free to close this down.

palob:
Actually all Numix Circle icons are already SVGcleaned.
Numix Square icons don't need to be before #3482.

Scour must be run with the options 
```
--disable-style-to-xml --enable-viewboxing --enable-comment-stripping
```
in order not to harm Inkscape editability. 

Generally speaking [``svgcleaner``](https://github.com/RazrFalcon/svgcleaner) is more robust and provides better cleaning as well as finer tuning options.

PR #3771, opened 2017-06-21
Surgeon Simulator

W-Floyd:
iconified_steam_icon_233720


palob:
You shouldn't do something like this:
no

The **S** as well as its shadow path should be separate paths on top of the red splash:
yes
yes2

W-Floyd:
Ah, I see. I'm busy tonight, so if you've done it already, go for it, otherwise I'll get to it tomorrow.

palob (approved):

PR #3776, opened 2017-06-23
qphotorec

W-Floyd:
iconified_qphotorec


palob:
How would it look with stripes of solid colours instead of a gradient?

W-Floyd:
I thought about it. I'll do that after I eat.

W-Floyd:
Like this? Loosely based the colours off some similar icons.
iconified_qphotorec
I tried a radial design, but it isn't really suited to this in the end.

palob:
Can you make the stripes overlap?
This will avoid the buggy outlines.
overlap

W-Floyd:
Ah ha, yup

W-Floyd:
I used straight lines, hopefully that will avoid any overhead that comes with curves :)
Still looks the same, so that's fine.
image

palob:
Curves are better in general because overlapping objects sharing the same outline like it is now also can shine through and cause those infamous jagged outlines. The overhead is definitely negligible.

In this case it's all fine anyway though.

palob (changes requested):
The Square symbol should go 1px down such that it is only 1px higher in comparison with Circle.

W-Floyd:
It already is 1px higher than the circle. Ordinarily, I would have done 2px higher than the circle, but 1px seemed to look best.

palob:
No, it is 2px higher currently.

W-Floyd:
Nevermind, I'm dumb. I fixed it, then forgot I fixed it :P
iconified_qphotorec

palob (approved):

PR #3784, opened 2017-06-27
Square icon for cdemu-client.

W-Floyd:
iconified_cdemu-client


palob (changes requested):
The shadows on the symbol itself need to be adjusted.

W-Floyd:
Ah, yup, didn't notice those. They're only 0.5px offset, is that standard for internal shadows?
iconified_cdemu-client

palob (approved):

palob:
No, there are 1/1px offsets for internal shadows as well. The .5px offsets were chosen where 1/1 caused too much of a 3D or height effect.

W-Floyd:
Okay, that's what I figured. Still working out what rules to go by :)

PR #3801, opened 2017-06-28
Uniform icons for Discord (Square & Circle)

Stefsinn:
| Original icon | Circle icons | Square icons |
| --- | :-- | :-- |

Updating the exisiting icons and adding a missing one for the following applications:

### Applications:
- [x] discord
- [x] discord-canary

### Details:
- [x] Icons for Numix Circle
- [x] Icons for Numix Square
- [x] Uniform design for all graphics
- [x] Using the original template files
- [x] Using the [original assets of Discord](https://discordapp.com/branding)


palob (changes requested):
+ Symbol size is too big, the symbol shouldn't exceed a width of 28px
+ The symbol's alignment with the pixel grid isn't ideal (isn't ideal with our current one as well).
+ In the canray icon the baseplate is misplaced, seems you shifted it accidently.
+ For the baseplate we use a vertical gradient instead of a solid colour, please cf. existing icons.
+ Please save the icons as *Optimised SVG*

palob:
Hi, thanks for contributing.
In addition to the comments above I'm not sure it's a good idea to directly use their SVG assets.

They state *if you're gonna use our art for something please keep it tasteful and send it our way for approval* which isn't compatible with our licence by any stretch. Maybe we should even alter our current icons anyway.

Stefsinn:
@palob Hi there!
Thanks for reviewing it so quickly. The icons were created for personal use in the first place. I will update my work according to the points you mentioned. 

Regarding the symbol size, I had the canary icon right next to it when making adjustments. I took it as reference and I had the feeling it's being too small compared to the available space. The next version will respect the limit of 28px.

PR #3772, opened 2017-06-21
Square calendar icons.

W-Floyd:
Day | Blue | Purple | Red
--- | --- | --- | ---
| 1 | iconified_calendar-blue-01 | iconified_calendar-purple-01 | iconified_calendar-red-01 
| 2 | iconified_calendar-blue-02 | iconified_calendar-purple-02 | iconified_calendar-red-02 
| 3 | iconified_calendar-blue-03 | iconified_calendar-purple-03 | iconified_calendar-red-03 
| 4 | iconified_calendar-blue-04 | iconified_calendar-purple-04 | iconified_calendar-red-04 
| 5 | iconified_calendar-blue-05 | iconified_calendar-purple-05 | iconified_calendar-red-05 
| 6 | iconified_calendar-blue-06 | iconified_calendar-purple-06 | iconified_calendar-red-06 
| 7 | iconified_calendar-blue-07 | iconified_calendar-purple-07 | iconified_calendar-red-07 
| 8 | iconified_calendar-blue-08 | iconified_calendar-purple-08 | iconified_calendar-red-08 
| 9 | iconified_calendar-blue-09 | iconified_calendar-purple-09 | iconified_calendar-red-09 
| 10 | iconified_calendar-blue-10 | iconified_calendar-purple-10 | iconified_calendar-red-10 
| 11 | iconified_calendar-blue-11 | iconified_calendar-purple-11 | iconified_calendar-red-11 
| 12 | iconified_calendar-blue-12 | iconified_calendar-purple-12 | iconified_calendar-red-12 
| 13 | iconified_calendar-blue-13 | iconified_calendar-purple-13 | iconified_calendar-red-13 
| 14 | iconified_calendar-blue-14 | iconified_calendar-purple-14 | iconified_calendar-red-14 
| 15 | iconified_calendar-blue-15 | iconified_calendar-purple-15 | iconified_calendar-red-15 
| 16 | iconified_calendar-blue-16 | iconified_calendar-purple-16 | iconified_calendar-red-16 
| 17 | iconified_calendar-blue-17 | iconified_calendar-purple-17 | iconified_calendar-red-17 
| 18 | iconified_calendar-blue-18 | iconified_calendar-purple-18 | iconified_calendar-red-18 
| 19 | iconified_calendar-blue-19 | iconified_calendar-purple-19 | iconified_calendar-red-19 
| 20 | iconified_calendar-blue-20 | iconified_calendar-purple-20 | iconified_calendar-red-20 
[...]

PR #3811, opened 2017-07-10
Some more distro icons

palob:
For #2183 
distros

distros-square


Foggalong (changes requested):
For the most part looking good!

* Voyager symbol is too big - it's supposed to be 24px for perfect squares
* Why the split baseplate on the arya icon?
* Use a light pink baseplate for pepermint

palob:
Revised:
distros2
distros2-sq

palob:
void

Foggalong:
@palob I legit fixed that locally and then pushed & deleted the wrong version. Silly me :angry: what settings are you using btw for saving as optimised SVGs? It looks like yours are coming about a bit leaner than mine

Foggalong (approved):

palob:
options as shown here #3772.

PR #3814, opened 2017-07-13
HTTrack square icon

mymindstorm:
| --- | :-- | :-- |

This is mostly just a cut and paste from the circle icon to the square template. Should help with #3481


palob (changes requested):
Thanks and welcome. The shadow is slightly misplaced. It should be just 1px further down relative to the symbol for Square icons.

mymindstorm:
By 1px further down relative to the symbol, do you mean 1px below where it already is or 1px below the symbol?

palob:
The latter, below the symbol. X coordinate should be the same as the symbol's one.
Please take a look at existing Square icons.

mymindstorm:
That commit should fix it @palob

palob (approved):
Thanks.

PR #3815, opened 2017-07-14
Lutris square icon

mymindstorm:
Comparison

Converted Lutris icon to square version


palob (changes requested):
The baseplate gradient is the wrong way round. It should be lighter on top.

mymindstorm:
@palob should be fixed

palob (approved):

PR #2656, opened 2015-11-12
Update README with Contributions section

ismaelgv:
Many people feel lost when they want to start contributing, this section could be helpful to newcomers.

This is just a draft. Feel free to make any change or correction. (I'm not a native English speaker).

PR #3818, opened 2017-07-16
Add contributions section to readme

### LINKED ISSUE #2647: How to contribute ?
I'd like to submit some icons instead of just requesting.

I have Inkscape, but I don't know where to start. Is there a template file somewhere that you have to start with ? Do I just modify an existing icon ?
Is there a list of guidelines to follow ?

Thanks :)
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28100990-how-to-contribute?utm_campaign=plugin&utm_content=tracker%2F514195&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F514195&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mymindstorm:
Continuing from #2656, fixes #2647.

@Foggalong Since you made a wiki page about contributing, should we just have it say "More info in the wiki" and link the wiki's contributing page?

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ This repository powers the generation of the all the Numix app icon themes accro
 Please report icon requests in this repo, providing all the details required. For normal applications follow [this video tutorial](https://plus.google.com/+NumixprojectOrg/posts/DkRmhFZuWez), for Steam games follow [this one](https://www.youtube.com/watch?v=BuUy4CzCoXc) and for Chrome apps just post a link to the webstore page. When filing your request please be respectful, patient, and remember that development is done solely on the back of donations.
 
 ### Contributions
-If you want to contribute to this project making new icons you must follow our [guidelines](https://github.com/numixproject/numix-wiki/wiki/Guidelines).
+If you want to contribute to this project making new icons you must follow our [guidelines](https://github.com/numixproject/numix-core/wiki/Guidelines).
 
 1. [Fork](https://help.github.com/articles/fork-a-repo/) our GitHub repository
 2. Create a new icon. You can use free open source tools like [Inkscape](https://inkscape.org/)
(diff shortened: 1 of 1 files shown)


Foggalong (changes requested):
@mymindstorm Aye, idea is to put the technical stuff in the wiki and just point people in that direction if they want to get involved.

It should probably go underneath "Hardcoded Icons" too because those two sections are linked.

mymindstorm:
@Foggalong first change was made, but I'm not sure about moving it below hardcoded icons. I think that "Icon Requests" and "Contributions" should be next to each other. What about moving hardcoded icons to the top if you still think it should be moved? 

On a side note, I don't really see what hardcoded icons has to do with the artwork section. Doesn't it just fix icons that aren't themed because an application didn't use `/usr/share/icons/`? If so, wouldn't it make more sense to give it it's own section?

mymindstorm:
Another idea: move "Hardcoded Icons" to the top of the "Artwork" section and rename it to "Unthemed Icons", and have it link [here](https://github.com/numixproject/numix-core/wiki/Unthemed-Icons).

Foggalong:
@mymindstorm it's currently under artwork because that's the part of the repo that it links to: the icons themselves rather than the script for generating the themes. The idea of it being next to Icon Requests is that you should be checking if the icon is hardcoded when you make a request.

The restructuring you've suggested makes sense, but I think it's more an indication that the README will have to be rewritten after the wiki has been updated so it better compliments the information there. I'll merge this for now and we can come back to further changes after that has been done :+1:

Foggalong (approved):

PR #3820, opened 2017-07-17
Fixes issues with square HP logo.

W-Floyd:
Before
iconified_hp_logo_old

After
iconified_hp_logo_new


palob:
This would be one for #3482.

W-Floyd:
Ah, thanks for checking that.

Issue #3484, opened 2016-11-07
Outdated Icons in Square


Foggalong:
There are some icons in the Square section which have been removed from Circle because they caused bugs or were just very outdated (think alternative versions of icons). I came across a few while getting the merge ready and no doubt more will be found as #3481 and #3482 are addressed.

andia89:
These are all the icons that are in square and not in circle. Not all of them are useless I would say
- [ ] 56B4_Battle.net Launcher.0.svg
- [ ] 5C12_World of Warcraft Launcher.0.svg
- [ ] aim.svg
- [ ] apport-old.svg
- [ ] battery.svg
- [ ] beatbox.svg
- [ ] bless-48x48.svg
- [ ] bluefish-alt.svg
- [ ] boot-repair.svg
- [ ] bt-logo.svg
- [ ] calc-alt.svg
- [ ] calc-old.svg
- [ ] calendar.svg
- [ ] celestia.svg
- [ ] chrome-namljbfbglehfnlonjmebceimaalofei-Default.svg
- [ ] chrome-oflhioojkbelepjlnafgmgkkjhojphcg-Default.svg
- [ ] chrome-onlgmecjpnejhfeofkgbfgnmdlipdejb-Default.svg
- [ ] config-users.svg
- [ ] cs-desklets.svg
- [ ] cs-overview.svg
- [ ] desktop-enviroment-gnome.svg
- [ ] desktop-enviroment-kde.svg
- [ ] desktop-enviroment-mate.svg
- [ ] desktop-enviroment-unity.svg
- [ ] distributor-logo-magiea.svg
- [ ] distributor-logo.svg
- [ ] filelight.svg
- [ ] file-manager-old.svg
- [ ] folder-sparkleshare.svg
- [ ] footnote.svg
- [ ] gnome-shell-extension-prefs.svg
- [ ] goa-account-facebook.svg
- [ ] goa-account-flickr.svg
- [ ] goa-account-google.svg
- [ ] goa-account-kerberos.svg
- [ ] goa-account-live.svg
- [ ] goa-account-owncloud.svg
- [ ] goa-account-pocket.svg
- [ ] goa-account-twitter.svg
- [ ] goa-account-yahoo.svg
- [ ] google-calendar.svg
- [ ] help.svg
- [ ] info.svg
- [ ] jabber.svg
- [ ] josm.svg
- [ ] krita.svg
- [ ] live.svg
- [ ] mahjongg.svg
[...]

PR #3827, opened 2017-07-20
Square cleanup

W-Floyd:
I have removed and renamed files in accordance with the circle theme, as well as optimized and fixed a few icons along the way.

Name | Icon
--- | ---
gnome-power-manager | iconified_gnome-power-manager
multimedia-audio-player | iconified_multimedia-audio-player
sparkleshare | iconified_sparkleshare
telegram | iconified_telegram
userinfo | iconified_userinfo
web-trello | iconified_web-trello
wordpress | iconified_wordpress


palob (commented):
This is good to go! Thank you @W-Floyd for your meticulous work here.
Is this all 100% in line with Circle and data?
By "optimization" you are referring to SVG restructuring, there is nothing in here for #3482, right?

W-Floyd:
@palob To the best of my knowledge, yes, it's in line with circle and data. Optimization is really resaving the file as an optimized SVG. I didn't check if each file need it though, I can can do that if you want.

Glad it mostly seems okay :)

palob:
No you needn't do it now. We can clean all files up script-wise after we'll have finished tackling our two big Square list issues.

palob (approved):

W-Floyd:
Okay, cool :D

W-Floyd:
Noticed that this seems to be in line with #3484, hope I did the right thing...

Foggalong (changes requested):
Just as an FYI, `aim.svg` was [moved to Numix base](https://github.com/numixproject/numix-icon-theme/blob/master/Numix/48/apps/aim.svg) because its design was specific to neither Circle or Square. There were a few others which made the journey with it iirc.

---

Bit out of the realm of this pull request, but you can add a commit to match the gradient of `distributor-logo-lubuntu` in Square to its Circle equivalent. The current one is a bit messed up.

---

`distributor-logo-magiea` and the data is still there, just by another name. When originally committed it was done so with a typo; the *i* and *e* were the wrong way around in [*Mageia*](https://en.wikipedia.org/wiki/Mageia) which has been corrected in the Circle file name and [data.json](https://github.com/numixproject/numix-core/blob/master/data.json#L3496-L3500). The file in Square just needs renaming to fix the mistake there too.

---

Not a question, but a comment: I remember the disagreements that @me4oslav and I had about the `preference-system.svg` design. He was the creator of both, but when the new (and current) design came in I was so reluctant to leave the old one behind. iirc that's what spawned the idea of keeping behind the older design behind with as `*-old.svg` files. It turned out not to be such a bad idea because we've ended up reusing bits of some of them elsewhere.

---

Relating to 1058702d1f0a60d58c4390a66e784a74cf2c1c97: when `thunderbird-old.svg` was deleted from Circle its gradient was moved to `thunderbird.svg` because it looked nicer than the grey one which was used before. Same needs doing for Square

| Circle | Square |
| --- | --- |
| screenshot_2017-07-21_17-38-25 | screenshot_2017-07-21_17-38-36 |

---

The Telegram in Square needs its gradient changed to match to that of Circle: from `#37aee2` & `#51c3f4` as stops to `#3db0e3` & `#50b8e6` for bottom and stop respectively.

palob:
By the way, the Mageia icon is one of those in #3482 which need to be updated with the Circle symbol.

palob:
This PR is alright Mageia-wise. @W-Floyd already did it for #3482 (assuming there wasn't one), it's on master with the right name.
The old yone can be deleted.

Foggalong:
@palob Good catch, I wasn't on the master branch when checking Mageia which is why I missed it. I think that's the only one I messed up for :grimacing:

palob:
Some removed old Circle icons are stored here https://github.com/numixproject/numix-labs/tree/master/Josh/Old-Icons/circle

W-Floyd:
As per request:
iconified_sparkleshare

W-Floyd:
iconified_distributor-logo-lubuntu

***

As for the *-old files that I removed, I see the point, but I think they've all gotten new-styled counterparts, correct? Personally, I would either move them somewhere else, or dig them out of git history when needed, but it's your call.

***

iconified_thunderbird
iconified_thunderbird-branded

***

iconified_telegram

Foggalong:
@W-Floyd @palob Sorry for the confusion, I wasn't saying we should keep the *-old files. I think most all of them are already in use or otherwise elsewhere look outdated at this point (looking at you apport-old). It was more a bit of Numix history :smile:

Foggalong (approved):

W-Floyd:
Ah, I see :)

Issue #3423, opened 2016-09-14
Icons for missing Nuvola Player 3 Web App Scripts


skyghis:
I see the issue  #1576 closed. great you find a way to support Nuvolaplayer integration scripts icons.
It would be great if all integration scripts have an icon.

Current status : 
![screenshot from 2016-09-14 10-19-33](https://cloud.githubusercontent.com/assets/1068552/18504919/fc1fd34e-7a64-11e6-8fde-c5789b889103.png)

I'm personally interested by the PlexMusic icon :innocent:

* [x] nuvolaplayer3
* [ ] nuvolaplayer3_8tracks
* [x] nuvolaplayer3_bandcamp
* [x] nuvolaplayer3_deezer
* [x] nuvolaplayer3_google_play_music
* [ ] nuvolaplayer3_groove
* [ ] nuvolaplayer3_jango
* [ ] nuvolaplayer3_mixcloud
* [ ] nuvolaplayer3_plex
* [x] nuvolaplayer3_soundcloud
* [ ] nuvolaplayer3_spotify
* [ ] nuvolaplayer3_yandex_music
* [ ] nuvolaplayer3_youtube

Foggalong:
We still need all the actual icon names to be able to fix the issue. The developer in the issue referenced mentions that the icons now follow a format, but without a list of names we still can't fix this.

bil-elmoussaoui:
This was fixed upstream, you can take a look at the issue reported on
nuvola's repo

Le 14 sept. 2016 12:33, "Joshua H. Fogg" <notifications@github.com> a
écrit :

Closed #3423 <https://github.com/numixproject/circle-core/issues/3423>.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/numixproject/circle-core/issues/3423#event-788924322>,
or mute the thread

Foggalong:
@bil-elmoussaoui See above comment. It's not that they're hardcoded, it's that we don't have the names.

skyghis:
@Foggalong The icons names are in hicolor theme. The exhaustive list in screenshot: 
![nuvolaplayer3_icons](https://cloud.githubusercontent.com/assets/1068552/18627034/f74fc728-7e58-11e6-9ccc-744fb8b92f8b.png)

Do you want me to create a separate issue per icon ?

Foggalong:
[...]

PR #3836, opened 2017-07-22
Few symlinks from #3423

bilelmoussaoui:
I wasn't sure if those symlinks are alright.


palob:
You mean you aren't sure this will work because of potential hardcoding?

bilelmoussaoui:
I wasn't sure if we should provide those icons here on Core(Circle/Square) or in Numix base

palob:
So far we did them here because they appear outside of Nuvola Player as usual app lauchers.
I think you can merge this here.

bilelmoussaoui:
Alright :+1:

PR #3841, opened 2017-07-22
Add symlink for msiexec

mymindstorm:
Seems to be a Fedora specific thing, also not sure if I'm doing this right.


bilelmoussaoui:
Thanks

PR #3842, opened 2017-07-22
Icon for winefile

mymindstorm:
image


mymindstorm:
Shoot, I didn't save one the right way

mymindstorm:
image

palob:
This is the our usual wine colours and our usual folder colour, right?
The contrast isn't ideal? Maybe the solution can be a greyish cabinet symbol like in the original icon.
See the one in Numix icon ``bareftp.svg``.

Please make the colours overlap like this:
overlap2

Foggalong (changes requested):
What @palob said. Nowhere near enough contrast between the yellow of the WINE baseplate and the yellow of the folder.

Switching it out for the grey filing cabinet sounds like a good compromise, except if I'm understanding correctly this is a file manager icon. If that's the case it should probably keep with the format of the rest of our [file manager icons](https://github.com/numixproject/numix-folders/tree/master/styles/6/Numix-Circle/48/apps). It'll need something other than the WINE baseplate to signify that it's a WINE app; possibly using [this icon](https://github.com/numixproject/numix-folders/blob/master/styles/6/Numix/48/places/default-folder-wine.svg) for the folder?

If you go through with this remember to submit a pull request with the same icon to [Numix Folders](https://github.com/numixproject/numix-folders/); if you don't fancy following through with versions for all the styles that's okay we can finish them off for you.

mymindstorm:
First of all, @palob I don't really get what you mean by making the colors overlap. I'm new to Inkscape, so I'm probably just be missing a menu somewhere, but is this what you meant?

image | image
--- | ---

I don't really think I'm even remotely close to what you meant since one fourth of it is always going to be above the others and the `wine.svg` icon looks like this:

image | image
--- | ---

If you could point me in the right direction, that would really help.

---

Color wise I think maybe using the brown or grey folder colors might work better? I had no idea that palette in [numixproject/numix-folders](https://github.com/numixproject/numix-folders/#colours) existed.

Brown | Grey
--- | ---
screenshot from 2017-07-24 15-54-17 |  image

---

Here is a prototype of the file cabinet version, but I don't see an easy way to integrate the wine logo thing into it:

image

Foggalong:
@mymindstorm I think you might have misunderstood; I was suggesting the WINE logo on the default yellow folder and a fairly standard (maybe red for wine?) baseplate. All the file managers are fairly uniform in design because they inherit the colours of the given style. See [here](https://github.com/numixproject/numix-folders#styles) for how each one looks

screenshot_2017-07-24_23-08-03

The only icon which doesn't follow this is the Numix Folders application itself because it's a) not a file manager and b) can be used to change the default colour folders have in the file manager.

screenshot_2017-07-24_23-09-36

The filing cabinet design is what @palob suggested and wouldn't need to incorporate the WINE symbol because it's already using the WINE baseplate, so the link to its mother program is already there.

palob:
@mymindstorm The overlapping is just a workaround for a SVG rendering bug which causes the background to shine through as ever so fine outline if it isn't done in some cases.

mymindstorm:
That makes a lot more sense to me, I'll start making the different styles and submit a pr to numix-folders.

image | image
--- | ---

palob (approved):

Foggalong (approved):

mymindstorm:
@palob @Foggalong numixproject/numix-folders#176

Issue #3701, opened 2017-04-29
Cleaning icon names in this repository


heyhippari:
As this is a developement repository, it might make more sense to have verbose names for some icons.

I'm thinking specifically about stuff like steam_icon_xxx or chrome-xyz123somethingorother.  
Naming them using a verbose name (For example `steam_icon_281990` would be `stellaris`) and then adding the proper icon name in data.json as a symlink would make editing down the line a lot easier.  
For example, it makes it easier to find if a game already has an icon when GOG or whatever releases a Linux version with another icon name.

We would obviously need to be done with #3481 and #3482 before, to not have to redo these lists as well, but I think it could be useful.

Here is a list of those icons as of April 29 2017:

- [ ] chrome-bikioccmkafdpakkkcpdbppfkghcmihk-Default.svg
- [ ] chrome-gebjgjhbjedcomcajgpodjgfjgkepgpl-Default.svg
- [ ] chrome-dkfhfaphfkopdgpbfkebjfcblcafcmpi-Default.svg
- [ ] chrome-epionbmgciknkbdibokmpmcfllkdfeeo-Default.svg
- [ ] chrome-koegeopamaoljbmhnfjbclbocehhgmkm-Default.svg
- [ ] chrome-mojepfklcankkmikonjlnidiooanmpbb-Default.svg
- [ ] chrome-cnciopoikihiagdjbjpnocolokfelagl-Default.svg
- [ ] chrome-ofbpedjdfkboehhknoohhenbpmgcjoik-Default.svg
- [ ] chrome-omimkinlomnncbmnceacpkmlbfaapojj-Default.svg
- [ ] chrome-pkbfakcphdokdnbdhlndlpcehebmplab-Default.svg
- [ ] chrome-gcjbnclfkeffmdfdhflnadahgmajfgcm-Default.svg
- [ ] chrome-eempgbpnkjnacmilmobpbhbfpdjdcpgd-Default.svg
- [ ] chrome-aeellliejpiclijnghognjagifplmmij-Default.svg
- [ ] chrome-iggpfpnahkgpnindfkdncknoldgnccdg-Default.svg
- [ ] chrome-knpkfcpnjfbniadmfchjpcigfhookhaa-Default.svg
- [ ] chrome-fljalecfjciodhpcledpamjachpmelml-Default.svg
- [ ] chrome-njkkjobcechefaoknodniidfjapgfoco-Default.svg
[...]

PR #3853, opened 2017-07-26
7BD8_halo.0: Adds icon for Halo (Wine app)

W-Floyd:
The game plays well under wine, love the nostalgia :)
iconified_7bd8_halo 0


bilelmoussaoui:
@W-Floyd @palob @Foggalong I think it's better to name the Wine & Chrome apps icons as `halo` (in this case) and a symlink to the real icon name. This will make it easier for us to find

W-Floyd:
Sounds like a good idea to me. Would the same apply to steam icons?

Just to confirm:
```
    "halo": {
        "linux": {
            "root": "7BD8_halo.0"
        }
    },
```

And, should there be a symlink (once [IO1](https://installation01.org/) is released, for example):
```
    "halo": {
        "linux": {
            "root": "7BD8_halo.0",
            "symlinks": [
                "installation-01"
            ]
        }
    },
```

Foggalong:
@bil-elmoussaoui @W-Floyd Yeah, I agree! That's why core was built so the linux root didn't have to match the icon name :)

palob:
I think it'd be cool to have descriptive names in core in the long term and get rid of the unwieldy ones.
Not only DB entries but base icon names as well.
Makes it easier to see what's covered.

Foggalong:
Base and in base icon theme? I don't follow

palob:
Basically this #3701

bilelmoussaoui:
@Foggalong @palob It would be even better if we can split the icons into different categories: Games, Web Apps, Browsers, Multimedia, Tools... This way, it will be much easier to find an icon (there's more than 2000 at the moment...) and things will get more complicated ones we merge in the Android icons in the future

Foggalong:
@bil-elmoussaoui Only problem I've been able to come up with doing that is some apps may fit into two categories; e.g a web app that's a game, or a multimedia tool

[Moka](https://github.com/snwh/moka-icon-theme/tree/master/src) sorts things by first letter (and then weirdly `web-*.svg`) but that again has the problem that you might not know what name an app is using.

Foggalong (approved):

PR #3857, opened 2017-07-26
add .editorconfig & markdownlint the README

bilelmoussaoui:
This PR adds a .editorconfig file, which will force indent size & type for the data.json. As you can see here in #3856 few lines didn't respect the indent size, this will fix it for all the cases where the contributor uses a text editor that supports editorconfig (Atom, VSCode, Gedit, Builder, Sublime Text....)

And also remove some unneeded empty lines on the README.md (there's now a project on Github that defines few rules for a cleaner and readable Markdown file)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+indent_size = 2
+indent_style = space
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.json]
+insert_final_newline = false
+
+[*.{py,json}]
+indent_size = 4
--- a/README.md
+++ b/README.md
@@ -1,25 +1,29 @@
 # Numix Core
-This repository powers the generation of the all the Numix app icon themes accross all platforms. This new method is designed to make keeping themes on different platforms on feature parity easier as well as making it as simple as possible to add support for new platforms. Licensed under the GPL-3.0+
 
+This repository powers the generation of the all the Numix app icon themes accross all platforms. This new method is designed to make keeping themes on different platforms on feature parity easier as well as making it as simple as possible to add support for new platforms. Licensed under the GPL-3.0+
 
 ## Artwork
 
 ### Icon Requests
+
 Please report icon requests in this repo, providing all the details required. For normal applications follow [this video tutorial](https://plus.google.com/+NumixprojectOrg/posts/DkRmhFZuWez), for Steam games follow [this one](https://www.youtube.com/watch?v=BuUy4CzCoXc) and for Chrome apps just post a link to the webstore page. When filing your request please be respectful, patient, and remember that development is done solely on the back of donations.
 
 ### Contributions
+
 We accept contributions, read [this page](https://github.com/numixproject/numix-core/wiki/Contributing) for more info.
 
 ### Hardcoded Icons
-To deal with hardcoded application icons Numix uses the [hardcode-fixer](https://github.com/Foggalong/hardcode-fixer) script. A list of the applications supported by the script can be found [here](https://github.com/Foggalong/hardcode-fixer/wiki/App-Support).
 
+To deal with hardcoded application icons Numix uses the [hardcode-fixer](https://github.com/Foggalong/hardcode-fixer) script. A list of the applications supported by the script can be found [here](https://github.com/Foggalong/hardcode-fixer/wiki/App-Support).
 
 ## Script
 
 ### Dependencies
+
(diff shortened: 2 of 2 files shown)


Foggalong:
>  remove some unneeded empty lines on the README.md

It looks like it's adding unneeded empty lines on my end?

bilelmoussaoui:
@Foggalong There must be a space btw a headline & the rest of the content

bilelmoussaoui:
Here's a list of the specifications https://github.com/markdownlint/markdownlint/blob/master/example/markdown_spec.md

Foggalong:
But why? There's no difference between how the two are rendered, and the extra lines looses the distinction between different header levels and the association of text with its header when viewed raw.

screenshot_2017-07-26_16-53-51

I'm fine with conforming to amarkdown spec provided the spec makes sense. I can't see and explanation of why the extra line between is needed and it's just making things less readable imo :confused:

bilelmoussaoui:
It's not about how things renders, but just to make those markdown files follow certains spec (the same as PEP8 for Python). Your code might looks readable even if you don't follow it.

And honestly, i don't really know the real reason why there should be an extra line in this specific case. i might open a new issue about it and get more information from the devs :+1: 

Anyway, it's just a suggestion i can revert that if you wish :)

Foggalong (changes requested):
On further inspection this doesn't look like a set spec anyways, making reference in the README that the linters default settings are just the author's preferred markdown style.

I think the idea of a markdown styleguide is interesting (and not something I'd thought about before) but there're enough weird things here that I'll pass on it.

.editorconfig
```diff
@@ -0,0 +1,15 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+indent_size = 2
```
Foggalong, line 6: Why is the indent size set to 2?

bilelmoussaoui: It's the default indent size for files other than .json & .py. xml files (SVG ones) exported using Inkscape are using 2 as indent size

Foggalong: Won't that also end up including things like the README, LICENCE, .gitignore, etc that it shouldn't?

bilelmoussaoui: I don't think anyone will ever modify the LICENSE file, .gitignore file shouldn't include any indentation. 
For the README, it depends on which indent size you think it should be set (for now we don't use it anyway)

Foggalong:
> the same as PEP8 for Python

Difference between this and PEP8 is that the latter is the official guide, written by the guy who created Python. For Markdown no such official spec exists, [by the looks of it](https://blog.codinghorror.com/standard-markdown-is-now-common-markdown/) because its creator is opposed to such a standard.

bilelmoussaoui:
I didn't knew about that. Well, i can revert that commit :+1:

Foggalong:
Me neither tbf, I'd never considered markdown specs before and only looked it up after seeing this. The [history](https://en.wikipedia.org/wiki/Markdown) of markdown and the efforts to standardise and extend it is a lot more interesting than I would have thought.

.editorconfig
Foggalong: That's a fair point; only place in README I can think of is a code indent and I can't think of a situation where we'd use one. It won't throw any errors/try to convert existing indentations will it? Thinking if someone accidentally opens the LICENSE file

bilelmoussaoui: @Foggalong It won't cause any thing unless you modify the file. Which should never happen, if the license should be changed the whole file should be replaced as i don't think we are going to ever write a license from scratch. 
The README file, doesn't use any indentation and i don't think we will have any complicated stuff to show on this repository.  
And, tbh i have been using 2 as an indent size for README files which looks really normal to me, as you never have something complicated there; it's almost as plain text file

Foggalong: Hehe, at the risk of taking this off topic twice on one thread we did actually have a Numix license for our closed source projects. It should in theory be somewhere in this repos git history because it was the license for the private Square repo before it was merged into here.

bilelmoussaoui:
@Foggalong So; what should be done here? Revert the README modifications, and keep the current .editorconfig? (we will need it as more tests will land => more code :P)

Foggalong:
@bil-elmoussaoui Yeah, just that. I'll be honest I'm not normally a fan of including things like `.editorconfig` files directly in repo masters, but given this is a repo for building packages it makes sense that there's gonna be some extra stuff floating around

bilelmoussaoui:
@Foggalong Done. You can merge whenever you want :+1:

Foggalong (approved):

PR #3860, opened 2017-07-29
Test svgo compression

bilelmoussaoui:
The svgo compression has a lot of plugins (see the .svgo.yml)
This PR shouldn't be merged at all right now, but it's more to discuss if we should use it or not.
I have run the script on all of the circle & square icons, the circle icons are still looking perfectly fine while a lot of the square ones look empty (at least on Nautilus. Maybe they're a part of #3482)
There's also two small issues (not caused by svgo): 
- a lot of icons have this viewBox `viewBox="0 0 48 48.00001" ` instead of this `viewBox="0 0 48 48" `
- there's a lot of header definitions either with only the viewBox or with the width & height or with an xmlns attr or without :P We should use the same one on all the icons (i guess with width & height, viewBox & xmlns)

We have approximately reduced the file size of the whole repository by 33%.


Foggalong:
If it's borking up the square icons then probably not until we've got around to verifying and correcting all of those. As a point for when this is actually proposed to be merged, keep the pull request adding the test and the one cleaning the icon separately; as is it's very big and difficult to verify.

bilelmoussaoui:
@Foggalong It's just testing the svgo implementation. I've made this PR more to discuss the possible plugins that should be enabled by default and which we should disable; The tests will definitely be added using a different PR

bilelmoussaoui:
It seems that Papirus guys use this too, here's their config file with some useful comments 
https://github.com/PapirusDevelopmentTeam/papirus-icon-theme/blob/master/tools/_svgo.yml
We might need to setup one for Apps icons and maybe a different one for the Base icon theme :+1:

palob:
Our [svgclean](https://github.com/palob/numix-tools/blob/7bbed7ad935c1f9943a1b2b2a2a74a74ef107823/numix-tools/svgclean) script is basically a config for [svgcleaner](https://github.com/RazrFalcon/svgcleaner). Might keep on hand some hints.

bilelmoussaoui:
@palob We can still stick with svgcleaner if you think it's better. SVGO has a lot of contributors and active development compared to svgcleaner (6 contributors compared to 75)

Foggalong:
@palob @bil-elmoussaoui I'll be honest, I'm super cautious about using either SVGO or svgcleaner after the problems using the latter has had in the past. Open to using either though if we can find a solution that works everywhere :+1:

mymindstorm:
@bil-elmoussaoui on the problem with `viewBox="0 0 48 48.00001"`: maybe you should run the files through a regex pattern like `(\d+\.\d{5,20})` and then round?

bilelmoussaoui:
@mymindstorm Yeah ofc :) we are just not sure which things to enable/disable on SVGO. There's also a way to do that using svgo without having to use any regex stuff :p

bilelmoussaoui:
Closing, I will open a new PR with scour instead of svgo

PR #3867, opened 2017-08-04
Few deepin symlinks

bilelmoussaoui:


palob:
Maybe link ``deepin-system-monitor`` to ``gnome-usage`` or ``stacer`` for a bit more diversity?

bilelmoussaoui:
Done :+1: 
There are still few icons used on the Dock but they don't have a .desktop file. And I'm not a deepin user to dig into their code for the right icon name :P

palob:
Can you please add those?
https://github.com/linuxdeepin/deepin-deb-installer/blob/master/deepin-deb-installer.desktop
https://github.com/linuxdeepin/deepin-draw/blob/master/deepin-draw.desktop

bilelmoussaoui:
To which ones you should i symlink them?

palob:
I'm not sure what *Deepin Draw* exactly is. So maybe to ``gnome-paint``, ``pinta``, ``latexdraw``, ``medit`` or something.

DEB Installer to ``synaptic`` (which I think it is the replacement for) or another package icon.

bilelmoussaoui:
@palob For Deepin deb installer, i think we better symlink it to a new icon for a new elementary app called eddy (see here both icons 
https://github.com/donadigo/eddy/blob/master/data/icons/64/com.github.donadigo.eddy.svg
https://github.com/linuxdeepin/deepin-deb-installer/blob/master/resources/images/deepin-deb-installer-128px.svg
)
What do you think about that?
The original deepin draw icon is https://github.com/linuxdeepin/deepin-draw/blob/master/logo/deepin-draw.svg so a symlink to gnome-paint would be good enough :)

Foggalong (changes requested):
Missing symlinks for Deepin Draw and Deepin Deb Installer

bilelmoussaoui:
@Foggalong I have added the deepin-draw symlink and I've proposed a better alternative other than a symlink for Deepin Deb Installer. What do you think about that?

palob:
So something like a package with a Debian swirl on it?

bilelmoussaoui:
@palob Yep :) which makes more sense

Foggalong:
@palob Huh, I thought we already had one of those

palob:
Well, we have got a Debian softwarecenter thingy.

bilelmoussaoui:
@palob @Foggalong This can be merged :) We can open a new issue for Deepin Deb Installer as there's no possible symlink. There's already a lot of icons symlinked to gdebi icon. 
A proper one for Deepin Deb Installer & Eddy (elementary OS app) will be nice ;)

PR #3873, opened 2017-08-07
Icon for Terminus

palob:
https://github.com/Eugeny/terminus/

128x128

terminus


icons/square/48/terminus.svg
```diff
@@ -0,0 +1,20 @@
+<svg width="48" height="48" version="1.1" viewBox="0 0 48 48" xmlns="http://www.w3.org/2000/svg">
```
bilelmoussaoui, line 1: Why there's a width & height property while on the circle one there isn't? I don't know if we should keep them as a lot of other icons have both. As far as I know, viewBox only is enough :) maybe the missing width & height can cause issues with Qt or another SVG rendering library

palob: That's how it is with all Circle icons. Probably because the Circle icons were treated with svgcleaner (and respective options) and new Square icons (well, new Circle ones as well) are optimised from Inkscape (using Scour). My Inkscape Scour options probably cause height/width variables not to come back.

Anyway, all icons show up properly with Qt, rsvg and Web SVG rendering engines.

Foggalong (changes requested):
Not that your icon is badly designed or anything, just something about it feels off for Numix. Maybe a shadow & light inconsistency or the ultra 3D feel? I don't know but anyways, I'd say to just switch this to our standard terminal icon with the Terminus colour scheme

palob:
terminus2

PR #3881, opened 2017-08-11
Icon and entry for korg (for editor via wine).

W-Floyd:
Not sure what's up with the name Wine gave the icon, but oh well. 

iconified_korg


palob (changes requested):
I'd prefer something like *korg* for the root icon name as well with the actual icon name as symlink.

W-Floyd:
~~Done.~~

palob (approved):

Foggalong:
> I'd prefer something like korg for the root icon name as well with the actual icon name as symlink.

@palob is there a reason for doing it that way?

palob:
Just so there isn't a file ``67F8_NewShortcut1_4E956BBDAB1B4156AEF14A08F107E9D7.0.svg`` in core but one with an actual name instead.

Foggalong:
You can have the file in core be `korg.svg` without adding the unused icon in circle with something like this:

```
"korg": {
    "linux": {
        "root": "67F8_NewShortcut1_4E956BBDAB1B4156AEF14A08F107E9D7.0
    }
},
```

EDIT: removed line with random `]`

palob:
Is that sole bracket intentional?

Foggalong:
@palob It was not, just my shody copypasting. Thanks, corrected the original comment!

W-Floyd:
So... should I revert my last commit? :stuck_out_tongue_closed_eyes:

Foggalong:
@W-Floyd yeah, didn't realise that was your original commit! Unless @palob has caught something I missed?

Foggalong (changes requested):
@W-Floyd yeah, didn't realise that was your original commit! Unless @palob has caught something I missed?

W-Floyd:
Do you want me to uncommit and force push, or do a proper revert?

Foggalong:
Uncommit probably the cleanest :+1:

W-Floyd:
Done :stuck_out_tongue_closed_eyes:

Foggalong (approved):

Foggalong:
@W-Floyd I've just come back to this issue... if it's a WINE app isn't it hardcoded? In which case we don't have to actually use the ridiculous icon name that they use by default

W-Floyd:
It is a wine app, so I guess? Not sure what you mean though, as it's already linking from `korg`, right?

Foggalong:
Just having an icon called `67F8_NewShortcut1_4E956BBDAB1B4156AEF14A08F107E9D7.0.svg` isn't sufficient to automatically apply the icon to the launcher. At least when I last used WINE the launcher that it creates uses a hardcoded icon path pointing to a specific icon location (e.g. `Icon=/path/to/korg-icon`) which needs to be changed to something like `Icon=korg` before it starts picking up icons from themes.

Admittedly it is a while since I've used WINE so someone can correct me if that's no longer the case

W-Floyd:
Mmm, just checked, so far as I can tell, wine doesn't hardcode the icon
path in, so the icon works correctly. It certainly worked with the Korg
icon at the time.

Foggalong:
Oh awesome, glad that they've fixed that!

W-Floyd:
@Foggalong Just checking back in on this (had to install the software again), the icon name is different now, I find it to be `62DF_KorgKontrolEditor._FF3416D6314D4FD48899393A083B0E82.0`

Should I submit a PR with the name changed (the only references to the old one are in relation to us), add the new name, or just drop this icon altogether?

As a side note, I'm sorry I've not made time to be active here at all (I still stay notified of things).
I figure I lost motivation once my square itch got scratched :stuck_out_tongue_closed_eyes:

Foggalong:
I'm guessing that WINE icons then change on a by user basis which puts it back into the annoying territory. Rather than a adding new symlink we'll just keep names generic for the approved WINE apps and treat them as hardcoded

Foggalong:
And don't worry about lack of activity! Everyones presence here comes and gos is waves, mine included

PR #3883, opened 2017-08-11
Icons and entry for Blueprint Tycoon.

W-Floyd:
Uses `blueprint_tycoon` for the file name, root `steam_icon_454060`.
iconified_blueprint_tycoon


palob (changes requested):
Design-wise nothing to be desired but can please you save the icons again? Seems optimisation is lacking something. There is lots of unneeded styles stuff for some paths.

W-Floyd:
Yeah, I keep noticing that happens. Any ideas on how to get rid of it? Saving it as optimized doesn't fix it, I've been editing files like these manually to remove those sections :/

palob:
Is this stuff already present in some way in the icons yours are based on?

Does running *File -> Clean up document* before saving anything?

palob:
In Inkscape preferences there is a *SVG output* section. Maybe you've checked the *Remove style properties* option (I haven't) which maybe sets all styles to *none*.

W-Floyd:
They're based on the provided templates, _Clean up document_ does not change anything, _Remove style_ is not checked. I'm using Inkscape 0.92.1 r

W-Floyd:
I went ahead and did it manually.

I also tweaked some of the spacing between the segments.
iconified_blueprint_tycoon

palob (approved):

PR #3888, opened 2017-08-14
Add Icons for Brackets-Electron

thenanobel:
@palob I've decreased the saturation of these icons. Are these okay?

| --- | :-- | :-- |
| Original Icon | Circle Icon | Square Icon |

### DIFF
--- a/icons/circle/48/brackets-electron.svg
+++ b/icons/circle/48/brackets-electron.svg
@@ -1,0 +1,137 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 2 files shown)


Foggalong:
Just as a heads up, you can push stuff to the same branch to update a pull request rather than having to open a new one :)

Foggalong (changes requested):
The contrast doesn't look too bad to me, but there's a slight problem with the gradient being used. The only variation between the top stop and the bottom stop should be a difference of 10L (in HSL).

i.e. if the bottom stop is `#ca450d` which is `hsl(13, 224, 108)` then rather than the top stop being `#f04a0f` which is `hsl(11, 225, 128)` it should be `#de4c0e` which is `hsl(13, 224, 118)`.

You also need to save the Inkscape files as "optimised SVG" rather than just SVG. The optimal settings can [be found here](https://github.com/numixproject/numix-core/pull/3772#issuecomment-310022946)

thenanobel:
Thanks. I've done everything. Sending a PR again. Hope, everything would be fine this time.

PR #3892, opened 2017-08-18
start working on #3701

bilelmoussaoui:
The whole thing was done automatically :P so i'm not sure if everything is perfect! 
I have used the mapping.json mentioned by @matemeszaros so a part of this commit is done thanks to his work :100:  

Once we are sure that we can use those new icon names i will provide a simple script to update #3482 automatically


bilelmoussaoui:
There's a Telegram_Web icon added with this commit. The icon is this one 
`chrome-clhhggbfdinjmjhajaheehoeibfljjno-Default.svg`
telegram_web
And it's for Telegram's Google Chrome app https://chrome.google.com/webstore/detail/telegram/clhhggbfdinjmjhajaheehoeibfljjno
We should really change this icon to the official one and rename this to something else and maybe use it for an other icon in the future.

bilelmoussaoui:
@Foggalong The any.do, web-ebay, airdroid & maybe other icons are on the Android PR too but as we those icons uses different icon names (chrome apps) we should maybe take care of this first to avoid fixing icons that already exists on Circle/Square :+1: 

One more reason to work on this is that we have a Slack icon named `slack` and there's a duplicate of it for the Chrome app `chrome-jeogkiiogjbmhklcnbgkdcjoioegiknm-Default`



Plus, we might need a convention for naming the original files. Should we go with 
- MixedCaps
- mixedCaps
- mixed_caps
- mixed-caps

Foggalong:
`mixed-caps` is my favourite followed by `MixedCaps`. `mixedCaps` is one of things in programming I cannot stand. No idea why :stuck_out_tongue:

bilelmoussaoui:
I would go for mixed-caps too

bilelmoussaoui:
Those icons will cause some conflicts
- steam_icon_60 -> game name is ricochet which already exists
- steam_icon_293780 the same as the one before

Foggalong:
@bil-elmoussaoui I'm not sure what you mean by the same as the one before?

bilelmoussaoui:
@Foggalong I mean the same issue as the one list just before `steam_icon_60`

bilelmoussaoui:
This PR is so much of a mess anyway. I will have to clean the things a bit :p

bilelmoussaoui:
I will create a new PR in few seconds.

PR #3894, opened 2017-08-19
Clean 8 steam icon names (not listed in #3701)

mrmeszaros:
Found 8 other icons not listed in issue #3701:
 - [x] steam_icon_115800 → owlboy
 - [x] steam_icon_233720 → surgeon-simulator-2013
 - [x] steam_icon_299660 → 6180-the-moon
 - [x] steam_icon_311690 → enter-the-Gungeon
 - [x] steam_icon_344850 → big-pharma
 - [x] steam_icon_346900 → ad-venture-capitalist
 - [x] steam_icon_355410 → radium
 - [x] steam_icon_474750 → reigns

PR #3896, opened 2017-08-21
rename icons to mixed-caps, remove duplicated icons

bilelmoussaoui:
This adresses #3701 
This PR does:
- Remove a duplicated whatsapp icon and replace it with a symlink
- Replace an old design icon of slack to a symlink, and keep the old one as `slack-alt`
- Replace an old design icon of telegram to a symlink and keep the old one as `telegram-alt`
- Replace a duplicated icon of `chrome-app-list` to a symlink
- Rename Steam & Chrome games/apps to mixed-name & update the data.json

List of missing icons on this PR : 
Few of this list are in #3894 the rest can be found easily using https://steamdb.info 

- [ ] `steam_icon_293780` : Can't be renamed  the game name is Crawl and there's already an icon with that name
- [ ] `steam_icon_60` : Can't be renamed  the game name is Ricochet and there's already an icon with that name


bilelmoussaoui:
@Foggalong This is now done :) there are still two missing icons other than #3894. I don't know what i should do with them :p

palob:
``steam_icon_60`` could be renamed to ``ricochet-game``. In the case of crawl I think both are games.

bilelmoussaoui:
The existing Crawl icon is not a steam game; I think we can easily fix that by splitting the icons into different folders. At least for games (only in the numix core repository). So we can have a games/ directory and inside we can put directories like /steam, /android, /wine

Foggalong:
As much as a hate this new FreeDesktop icon name format, maybe go for something like `crawl.dev1.svg` and `crawl.dev2.svg`, where `dev1` and `dev2` are the two developer names?

bilelmoussaoui:
I will try to clean this PR up and resolve the merge conflicts as soon as possible. @Foggalong What do you suggest as an icon name for both Crawl icons?

Foggalong:
@bil-elmoussaoui See above

bilelmoussaoui:
I'm going to close this and create a new PR, those conflicts on data.json are really hard to fix manually :p

PR #3900, opened 2017-08-25
Icons for Tranqil

palob:
https://github.com/nick92/tranqil

tranqil-ori

tranqil


Foggalong (approved):
The contrast of the white is a nice addition not in the original icon :+1:

PR #3907, opened 2017-09-04
Icons for Frinika

palob:
https://sourceforge.net/projects/frinika/

frinika

frinika


Foggalong (commented):
How would this look with maybe a shade of pink for the music note? Just thinking of adding some more colour into the design

palob:
Pushed this
frinika

Foggalong (approved):

PR #3909, opened 2017-09-07
More Flatpak symlinks

palob:


Foggalong:
I'm so glad that Flatpack, Snap, etc didn't hardcode these icons otherwise this would have created a Nightmare.

Foggalong (approved):

PR #3912, opened 2017-09-13
Added icons for Sublime Text 3.0 Build 3143

jaimelive:
| --- | :-- | :-- |
|  original | circle | square |

## Application:
Sublime Text 3.0 - Build 3143

## Sources:
From the official logo: official_logo


bilelmoussaoui:
Wouldn't be wiser to keep the old logo for sublime text 2 and use the new one for the latest version?
Btw, why the file permissions of both icons were changed from 644 to 755?

jaimelive:
@bil-elmoussaoui Indeed it is.
The permissions change was a SMB induced mistake, now corrected.

bilelmoussaoui:
If you can rename the icon to sublime-text-3 that would be awesome.  How did you create the icons?using Inkscape? The icons must be saved as optimized

jaimelive:
@bil-elmoussaoui Done and done.

palob:
HI, thanks for contributing.

I'm not sure about the gradient but a plain shadow looks somewhat strange (and isn't visible when zooming out) or doesn't it?
subl2

jaimelive:
@palob In what way? I like it. Either way I followed the [guidelines](https://github.com/numixproject/numix-core/wiki/Guidelines)

> The symbol shadow is a duplicate of the symbol background coloured black, made 10% opaque and moved 1px to the right and bottom and of course moved below the symbols. If the symbol shadow consists of more then one object, merge them into one single path.

palob:
Usually we try to avoid gradients on symbols but there's exceptions proving the rule.

palob:
@jaimelive Can you please add a database entry for the new icon to ``data.json``?

jaimelive:
@palob I had misunderstood you, I thought you were referring to the symbol shadow.
Would something like this be better?
sample

Foggalong (changes requested):
Yeah, a solid shadow rather than a gradient one would be preferable to that it fits in with the rest of the theme

mrmeszaros:
I was wondering if it would look even better with a shadow on the upper segment too.
@palob what do you think?

Foggalong:
@matemeszaros There isn't such a shadow on the original icon though?

mrmeszaros:
@Foggalong no, there isn't. But since the symbol is very ribbon like, it may look better with the shadow that emphasizes this.

Foggalong:
@matemeszaros Usually we try and stick with something close to what upstream uses unless it looks particularly bad. Maybe it would look significantly better, I'd just not though it was missing until you pointed it out so feel free to have a play with it and see how it turns out

jaimelive:
@matemeszaros I'm having trouble visualizing your suggestion. Due to the "position of the light source" the shadow is projected 1px to the right and 1px to down, so it wouldn't normally cast a shadow on the upper segment. Can you do a quick sketch of what you mean?

mrmeszaros:
@jaimelive I have been trying to come up with a solution with the shadow I mentioned, but did not find any that looks well enough. I say forget what I suggested. Thanks for the effort :)

jaimelive:
@Foggalong @palob @matemeszaros 
I removed gradients and tried to make the ribbon more explicit by changing the color of each segment.
Hasn't been committed yet.
small_sample

sample

large_sample

palob:
Yay, that's the way to go.

Foggalong:
I love that solution, great work @jaimelive!

jaimelive:
@Foggalong Sorry for the delay, changes commited.

Foggalong (approved):

Foggalong:
We'll have to keep an eye on this for if they rename the icon lines to `Icon=Sublime` for version 3, then we'll have to switch them around.

PR #3920, opened 2017-09-15
add scour optimizations tests

bilelmoussaoui:
This PR introduces a new test. The files now must be optimized using scour with those options : 

```shell
scour -i input.svg -o output.svg --disable-style-to-xml --create-groups --renderer-workaround --strip-xml-prolog --enable-comment-stripping --remove-descriptive-elements --enable-id-stripping --protect-ids-noninkscape
```
Those options are the same as the ones described on the wiki. 

The tests, will make sure that the icons and the templates are optimized correctly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,8 @@ before_install:
     - sudo apt-get install -y python3-pip python3-gi python3-gobject gir1.2-gtk-3.0 libgtk-3-dev libgirepository1.0-dev
 install:
     - pip3 install --upgrade pip
-    - pip3 install cairosvg pep8 jsonschema
+    - pip3 install cairosvg pep8 jsonschema scour
 script:
     - sh ./tests/pep8.sh
     - python3 ./tests/schema.py
+    - python3 ./tests/scour.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,4 +14,4 @@ install:
 script:
     - sh ./tests/pep8.sh
     - python3 ./tests/schema.py
-    - python3 ./tests/scour.py
+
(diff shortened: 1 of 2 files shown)


Foggalong (changes requested):
Needs PEPing. Also, is there a way of testing this before merging like we did with Python?

bilelmoussaoui:
Yeah, i will take care of the PEPing. For testing this, you can easily optimize few icons from circle (the icons are ordered alphabetically) so you can take 0ad.svg for example, optimize it using Inkscape or scour, and then just run python3 ./tests/scour.py from the numix-core directory. And you should see that the 0ad icon is optimized correctly 👍 
For now, this shouldn't be merged as the tests verifies all the icons which will take a huge time each time... So we should optimize everything first and then i will switch to verifying the latest icons added on the latest commit or something like that

bilelmoussaoui:
I have disabled the scour tests at the moment, will look into this once the optimized icons are on master

bilelmoussaoui:
As scour doesn't provide the same output everytime you run it, we will have issues with this kind of tests.
But we can implement our kind of tests like: 
- Verify if there's any comment
- Verify if there's any meta
- Verify if there's an empty group
and we can keep adding tests in order to get the things done

Foggalong:
@bil-elmoussaoui I suppose that's preferable to nothing at all :) also it'd be good to have tests to make sure in a PR there's the proper changes to the data.json, and that a Circle and Square icon are both added

bilelmoussaoui:
This can be closed too, as we can't test if the icons were optimized using scour. The output will be different each time :( 
I might try to implement something manually in the future

PR #3921, opened 2017-09-15
use scour to optimize all the icons

bilelmoussaoui:
This PR uses scour to optimize all the icons with the same options as described in the wiki (with Inkscape)


bilelmoussaoui:
This must be merged before #3920, so we can test if the current tests will always pass in the future

Foggalong:
dat diff.

bilelmoussaoui:
It saves almost 5mo :p

Foggalong:
Wow, 5 whole months in just on merge? That's some bang for your buck right there :wink:

bilelmoussaoui:
Haha, I meant 5 megabytes :p

palob:
There's actually no need to run this on Circle as I think I've checked that all icons already have been cleaned with svgleaner.

I wonder where those 5mb come from. Can you check? Circle or Square?
Many Square icons have been cleaned before as well, with bad options which is where the one-liner files originate from.

Anyway we need to scrutinise visually.

bilelmoussaoui:
The reason why I re-run all the icons using scour, is to have a standard optimizations method so we can set up a test to validate the icons. Otherwise, icons optimized with svgcleaner will be marked as "unoptimized"...

A lot of icons in Square were optimized with bad options, visually looking on Nautilus, everything seems to be fine 👍 

I will take a look where the 5mb size diff comes from exactly. But if you take only circle/0ad.svg and re-optimize it using Inkscape, there's a +14/-38 diff

Foggalong:
@bil-elmoussaoui Any joy with this?

Foggalong (changes requested):
@bil-elmoussaoui Any joy with this?

bilelmoussaoui:
@Foggalong what i should do?

Foggalong:
@bil-elmoussaoui you'd said you were going to look into where the diff came from as a way of checking it was all good :P but yeah I'm fine to merge this as long as the merge conflicts are addressed

bilelmoussaoui:
@Foggalong i will take care of this evening and provide more information about that huge diff

Foggalong:
@bil-elmoussaoui Any joy? Cautious of leaving this issue for too long before having it merged because the conflicts are only going to grow :disappointed:

bilelmoussaoui:
The easiest way to fix this is just starting from master again, i'm afraid to break any changes (during the conflict's resolution:/)
The script to optimize the icons is running right now! I will create a PR once it's done; shouldn't take a lot of time (max 10minutes)
@Foggalong Can you please also take a look at #3896? Otherwise, the diff will grow a lot there too :(

Foggalong:
@bil-elmoussaoui Do you want me to do #3896 then your upcoming scour pull or wait for the scour pull then look at #3896?

bilelmoussaoui:
@Foggalong Let's take care of the scour part first :+1: 
The other PR can't be merged anytime soon anyway, there are two games that have the same name. i don't know how to deal with that

PR #3922, opened 2017-09-16
Icon for web IRC client, The Lounge.

W-Floyd:
Not sure if this is the right naming scheme, I did this for an Epiphany web app icon, though I had to change the .desktop file myself.

iconified_lounge


palob:
Yeah, unfortunately Epiphany web app launchers have the icon paths hardcoded by default. I think this should be reported as bug.

palob (approved):

W-Floyd:
A step in the right direction might be for them to allow entering an icon name during setup. I'll need to look into that...

Foggalong (approved):

PR #3939, opened 2017-10-11
LibreOffice Writer circle icon consistent with oficial branding

leojimenezcr:
LibreOffice Writer circle icon consistent with [look and feel of oficial branding](https://wiki.documentfoundation.org/Visual_Elements). So people will recognize it more easily.

The icon is based on libreoffice-main.svg file and mixed with [oficial graphical elements and color pallete](https://wiki.documentfoundation.org/File:LibreOffice_Initial_Icons-pre_final.svg).

proposal-libreoffice-writer

### DIFF
--- a/icons/circle/48/libreoffice-writer.svg
+++ b/icons/circle/48/libreoffice-writer.svg
@@ -1,46 +1,298 @@
-<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 48 48">
- <defs>
-  <linearGradient id="linearGradient3764" x1="1" x2="47" gradientUnits="userSpaceOnUse" gradientTransform="matrix(0,-1,1,0,-1.5e-6,47.999998)">
-   <stop style="stop-color:#3aa0b2;stop-opacity:1"/>
-   <stop offset="1" style="stop-color:#40acc1;stop-opacity:1"/>
-  </linearGradient>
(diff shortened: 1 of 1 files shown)


palob (changes requested):
Please take a look at our design gudelines and existing icons.

leojimenezcr:
Hi. I think, according to design guidelines, I need to check the size of the symbol and add a shadow to it. I'm right? Any other observation?

palob:
It is crucial that all elements be aligned with the pixel grid, otherwise the icon looks blurry.
Usually we don't use gradients in symbol elements.
Some of your icons are too fine grained to be recognisable at smaller sizes, the Math one in particular.

Some of your icons use a stroked style which is distinctly un-Numixy.

I think for all icons you took the symbol right away from the original one. This defeats the purpose of a custom icon theme (and in other cases might not be compliant with our licence).

I don't think the way the dog-ear thing overlaps the symbol works out at all. Please find a better way of incorporating it.

leojimenezcr:
Thanks for take time to response. I will work in new propose.

palob:
Thanks, the amont of criticism might sound heavy, so I'm glad you're still enthusiastic.

W-Floyd:
Maybe not have gradients except the baseplate?

Foggalong (changes requested):
Extra extra note as well: these design changes will also have to be carried across to the Square theme.

PR #3945, opened 2017-10-11
LibreOffice main circle icon consistent with oficial branding

leojimenezcr:
I changed a little bit the LibreOffice main circle icon. I applied a lighter gray, according to the official color palette

proposal-libreoffice-main


palob (changes requested):
Please take a look at our design gudelines and existing icons.

Foggalong (changes requested):
As an extra note, the icons need to be saved as optimised SVGs before we merge them to reduce repo and package sizes.

PR #3969, opened 2017-11-05
Icons and entry for Exherbo.

W-Floyd:
For #2183

I think their logo sucks to begin with, but I'm not the one who has to live with it :stuck_out_tongue_closed_eyes: 

iconified_distributor-logo-exherbo


W-Floyd:
I notice that optimisation makes no difference to [these extra bits of styling](https://github.com/numixproject/numix-core/pull/3969/files#diff-f30b2135c2e6bea111ec9ac91c288123R14). If I ever save as a regular Inkscape file, then I have to manually remove them if I want them gone. Any ideas? Manually using scour makes no difference.

palob (commented):
How does it look without the outer stroke?
Generally Numix Circle/Square icons don't use full-blown black or white but something like L36/L249 at the darkest brightest.

Can you simplify the stacking?
stacking

W-Floyd:
Ah yeah, my bad on the black.

So stacking like this?

image

And for trying without the border, I think a darker baseplate might work with it. Thoughts?

image


image

palob:
Yes, stacking like this for option 1. Option 2 might work with a blackish baseplate.

W-Floyd:
Okay, so the question now, do we want to go without the black border?

iconified_distributor-logo-exherbo

Foggalong (commented):
Yeah, without the black outline in my opinion. Makes a really shit logo look less shit

W-Floyd:
lol Like lipstick on a cow... I'll commit that soon.

Foggalong (approved):

PR #3973, opened 2017-11-06
Add mkchromecast icon

W-Floyd:
iconified_mkchromecast

I have based the design from the Google provided Chromecast symbol, not the mkchromecast provided symbol, so it's slightly different.


palob (commented):
Symbol is fine but there is this cruft in the source again.

W-Floyd:
ugh, right. Just remove it manually again then?

palob:
Yep, it doesn't look broken afterwards or does it?

W-Floyd:
Nah, it's always fine, so long as the opacity and colour is retained.

palob (approved):

Foggalong (approved):

PR #3988, opened 2017-11-22
add db entry existence verification

bilelmoussaoui:
This PR add's a new test to verify if every item in the database has an icon in both themes circle and square :)
More tests are coming!


bilelmoussaoui:
```
 The icon rpcs3 doesn't exist in the theme circle 
 The icon rpcs3 doesn't exist in the theme square 
 The icon plague2 doesn't exist in the theme circle 
 The icon plague2 doesn't exist in the theme square 
```
Some of the icons don't exist in both circle and square but they're in the data file

Foggalong:
@bil-elmoussaoui Does that need resolving before this can be merged?

* [Commit](https://github.com/palob/numix-core/commit/a7d8b9985ac6cf9640d9a9cd4ebe9e534f513967) which added rpcs3
* Haven't been able to find where plague2 is coming from

bilelmoussaoui:
The commit added rpcs3 as a root icon name but it should be rcps3

Foggalong:
Similarly, fixing this in the same PR is fine since it directly relates to this issue

bilelmoussaoui:
@Foggalong this can be merged now

Foggalong (approved):

PR #3990, opened 2017-11-23
add icon entry existence verification

bilelmoussaoui:
This test adds a verification to whether the icon has an entry in the database file

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,4 @@ install:
 script:
     - sh ./tests/pep8.sh
     - python3 ./tests/schema.py
+    - python3 ./tests/icons_entry.py
--- a/data.json
+++ b/data.json
@@ -3018,6 +3018,11 @@
             "root": "cs-cat-admin"
         }
     },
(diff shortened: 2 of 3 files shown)


bilelmoussaoui:
The test fails because of those missing icons entry in the database
```
 The icon highlight doesn't have any entry in the database 
 The icon cs-desktop doesn't have any entry in the database 
 The icon monogame doesn't have any entry in the database 
 The icon distributor-logo-void doesn't have any entry in the database 
 The icon rcps3 doesn't have any entry in the database 
```

Foggalong:
@bil-elmoussaoui Can you make a pr to fix these?

bilelmoussaoui:
In the same PR or a new one?

Foggalong:
In the same PR is fine since it directly relates to this issue

bilelmoussaoui:
@Foggalong I have fixed both tests on the same, are they're somehow related :)

Foggalong (approved):

Issue #3889, opened 2017-08-14
Icon for Piper


rautesamtr:
| Application name | Icon name | Desktop file name | Original icon |
| --- | :-- | :-- | :-- |
| Piper (exec: piper)| org.freedesktop.Piper | org.freedesktop.Piper.desktop | ![original](https://cdn.rawgit.com/libratbag/piper/master/data/org.freedesktop.Piper.svg) |

Description: Piper is a GTK+ application to configure gaming mice, using libratbag via ratbagd.

Website: https://github.com/libratbag/piper

bilelmoussaoui:
Fixed

PR #4008, opened 2017-12-13
Add icons and entry for Piper.

W-Floyd:
For #3889

Hope it's okay that I played with the perspective for the square symbol. It didn't feel right to keep it the same as the circle but change the shadows.

iconified_piper


palob (commented):
Cool design but looks a bit busy I'm afraid. Is it recognisable even at 48px?
I think the baseplate should get a more distinctly contrasting colour.

W-Floyd:
So-so, I guess. Any ideas?

iconified_piper

palob:
I haven't come up with a better idea either so far which is why I've chickened out of doing this icon.

W-Floyd:
I see... I don't see other icon packs that have made one either.

Foggalong (approved):
It's not perfect, but better than nothing. We can come back to it later if needed

PR #4012, opened 2017-12-16
distributor-logo-gnu - GNU icons and entry.

W-Floyd:
For #2183

To preempt the "at 16/32/48px?" question, it is okay at 48+, which in this instance is probably fine.

iconified_distributor-logo-gnu

Based on [this](https://www.gnu.org/graphics/gnu_hornedword.svg) GNU logo.


palob (approved):
Definitely fine this way, just not as cute as a real gnu.

W-Floyd:
I wanted to do the regular logo, but it is too fine-grained to work, I think. I'm happy enough with this, though.

Foggalong (approved):

Issue #3971, opened 2017-11-05
Icon for Meshlab


bkanuka:
| Application name | Icon name | Desktop file name | Original icon |
| --- | :-- | :-- | :-- |
|Meshlab| `/usr/share/pixmaps/meshlab_32x32.xpm` | `meshlab.desktop` |![icon](http://www.meshlab.net/img/meshlabLogo.png)|

Description: the open source system for processing and editing 3D triangular meshes.
It provides a set of tools for editing, cleaning, healing, inspecting, rendering, texturing and converting meshes. It offers features for processing raw data produced by 3D digitization tools/devices and for preparing models for 3D printing.

Website: http://www.meshlab.net/

Yes, I know it's hardcoded, but it's not difficult to fix if an icon existed.

bkanuka:
original icon svg is here: https://github.com/cnr-isti-vclab/meshlab/blob/master/src/meshlab/images/eye.svg

I have done my best simplifying the icon. I think if this was put over a gray background, it would look nice and keep the "feeling" of the original icon. 
[drawing.svg.zip](https://github.com/numixproject/numix-core/files/1444174/drawing.svg.zip)

bkanuka:
I took my best shot at an icon. It is very much based on the original icon svg. I'm really hoping someone can complete the last few steps. I'm *really* not sure what I'm doing!
[meshlab.svg.zip](https://github.com/numixproject/numix-core/files/1444366/meshlab.svg.zip)

bilelmoussaoui:
Fixed

PR #4030, opened 2017-12-19
meshlab - Icons and entry for Meshlab.

W-Floyd:
For #3971

I tried doing an eye thing too, but it really wasn't looking much chop, nor does it signify the nature of the software. As such, I think something to do with a mesh (in this case, triangulation) is more appropriate, and easier to pull off in this theme.

iconified_meshlab


palob:
Does this work with this crosshair/grid background used on some icons?

W-Floyd:
Say, that's an idea... I'll try that later.

W-Floyd:
iconified_meshlab

palob (approved):

Foggalong (approved):

bkanuka:
Thanks for your work on this @W-Floyd ! 
I really like the icon you drew (and the idea of not using the eye), but I don't think the grid improves the design. I think it just adds noise to a theme that's otherwise pretty minimal. Plus a grid isn't even a common tool in Meshlab.

Anyway, really just glad to get this icon covered by numix - so thanks again

W-Floyd:
Hmm, I see what you mean... @palob you still think grid, or no?

palob:
I think it's a bit bland with just a white background. Maybe a light green background looks OK.

Foggalong:
Yeah it definitely needs something just beyond the plain white baseplate.

Issue #2294, opened 2015-06-18
Missing app icons in kde


andia89:
Looking at [#580](https://github.com/numixproject/numix-icon-theme/issues/580) I recognized that a a few icons are actually app icons and not base icons. So I deleted them from there and consequentely open a new pull request with the missing icons here:
- [x] Charm
- [x] esd
- [x] kblogger
- [x] kbruch
- [x] kchart
- [x] kformula
- [x] kivio
- [x] kjournal
- [x] korgac
- [x] kplato
- [x] kpresenter
- [x] ksniffer
- [x] kspread
- [x] ktip
- [x] kverbos
- [x] kword
- [x] plasmagik
- [x] preferences-contact-list
- [x] preferences-desktop-cryptography
- [x] preferences-desktop-display-color
- [x] preferences-desktop-filetype-association
- [x] preferences-desktop-gaming
- [x] preferences-desktop-icons
- [x] preferences-desktop-mouse
- [x] preferences-desktop-printer
- [x] preferences-desktop-text-to-speech
- [x] preferences-plugin-script
- [x] preferences-system-cron
- [x] preferences-system-windows-move
- [x] preferences-web-browser-adblock
- [x] preferences-web-browser-cache
- [ ] preferences-web-browser-identification
- [x] preferences-web-browser-shortcuts
- [x] semn
- [x] strigi
- [x] utilities-desktop-extra

![unbenannt](https://cloud.githubusercontent.com/assets/6475757/8229404/b87056f0-15b6-11e5-8bc2-f5810b399a93.png)
##

palob:
Maybe you can take off kchart, kformula, kivio, kplato, kpresenter, kspread and kwords. 

Those are applications from the former KOffice suite which is dead for 4 years now and succeeded by the Calliga Suite.
If needed nonetheless the icons could be done as symlinks to the Calligra icons.

Foggalong:
Symlinks would be fine yeah. I wouldn't put it past someone out there to be still using the old office. We've had weirder things

W-Floyd:
Checking through this list:
* `kbruch` has had its own icon for a while.
* `preferences-desktop-gaming` is already symlinked from `applications-games`

I'll prod at some of the others.

W-Floyd:
`ksniffer` has not been updated in [~7~ 8 years](http://ksniffer.sourceforge.net/index.php?section=home). If anyone has ideas on a good symlink, that'd be great, but I'm coming up short on a generic icon for it.

[...]

PR #4042, opened 2017-12-22
kverbos - Icons and entry for kverbos.

W-Floyd:
#2294

`kverbos` has not been updated, as far as I can tell, in [9 years](https://kverbos.en.uptodown.com/ubuntu). Only reason I did this was because it's so simple, and that I would be interested in seeing more flags have icons (if it would be useful).

iconified_kverbos


palob (approved):

Foggalong (changes requested):
Can you add a drop shadow under the top red segment so it looks as if the yellow is slightly inset?

W-Floyd:
Oh come off it Travis...

iconified_kverbos

bilelmoussaoui:
@W-Floyd It seems that a PR broke the data.json file :(

Foggalong (approved):

palob:
The yellow band isn't correctly aligned but I can't fix it right now.

W-Floyd:
This should be my fix :/

Edit: Nevermind, I see it's been done already :laughing:

PR #4054, opened 2017-12-24
Integrate Legacy Themes

Foggalong:
I've got a christmas surprise for Numix users; I'm switching our two main legacy themes (uTouch and Shine) over to the Core build system. This means that even if nobody is actively adding new icons to the theme they can still get symlink updates.

It's also gonna mean that, at least for now, the dedicated [numix-icon-theme-shine](https://github.com/numixproject/numix-icon-theme-shine) and [numix-icon-theme-utouch](https://github.com/numixproject/numix-icon-theme-utouch) repos are gonna go away, and anyone wanting to build packages should do it directly from here. In terms of other changes to the rest of the Core workflow:
* When adding new icons, you won't be expected to add ones for Shine and uTouch too
* Renames of icon SVGs must be done in Shine & uTouch too to prevent breakage
* We won't be accepting issues which are exclusively about Shine or uTouch
* We will be accepting pull requests which bring Shine or uTouch more up to date

This won't keep either of these themes up to date like Circle or Square and you should definitely use them if you're wanting a theme, but if you do insist on using Shine or uTouch they won't be as out of date as they would have otherwise been. So far I've only pushed Shine, but uTouch will follow shortly and an announcement about it will be made to the community over the next few days :smile:


bilelmoussaoui:
Regarding the packaging of the icons, as I have mentioned before in the Travis integration issue; we could easily (I guess) auto build the icon themes and push them to the right repositories automatically each two weeks for example and create a new release with just the diff of the two versions without having to do that manually. This will allow all the themes to be up to date without having to spend few minutes to do that manually

palob:
Does this build, i.e. are all icon names correct w.r.t. the the latest ``data.json``?

palob (approved):

Foggalong:
@palob Yeah; it wouldn't if it had been a straight merge but in 0de2a69 and 935d60b I went through renaming the old icons in Shine so that they matched up with the current names in Circle & Square. I'll be doing the same with uTouch today :)

PR #4055, opened 2017-12-24
stop replacing _ with a dot

bilelmoussaoui:
Few android icons has _ in their icon name. And as we don't need to keep it that way, the icons can be added directly with their correct icon name on the data file

### DIFF
--- a/gen.py
+++ b/gen.py
@@ -122,7 +122,7 @@ if platform == "android":
         for output_name in icon.get("android", []):
             source = "icons/{0}/48/{1}.svg".format(theme, format(icon_name))
             output = "{0}/{1}.png".format(android_dir,
-                                          output_name.replace("_", "."))
+                                          output_name)
             if path.exists(source):
                 convert_svg2png(source, output, 192, 192)
 elif platform == "linux":


Foggalong:
I believe the reason for this was that Android doesn't allow for `_` in the icon titles. @paolorotolo might be able to confirm

bilelmoussaoui:
The firefox beta uses it in the icon name, unless I didn't understood how everything works :p

bilelmoussaoui:
Ah okey!  I have completely misunderstood how things works. Currently, all the android entries has the packageName as an icon name. 

What we would like to do, is to have both, the packageName and the ComponentInfo. So we can generate the appfilter.xml automatically too. 

So this PR should be closed, but there's still something that should be done to fix this for the android gen. Otherwise, we will have to do it manually :(

I'm thinking of two ways to do this:
1 - We add the ComponentInfo to the data.json file, which will make things more complicated to handle :( 
2 - We add the appfilter.xml to this repository, and we only add the output icon name to the data.json file, the gen script can easily then check this file and if one icon doesn't exists the generated appfilter.xml will only contain the icons that were already added.

What do you think?

bilelmoussaoui:
The appfilter.xml content is something like this 
```
<item component="ComponentInfo{com.opera.mini.android/com.opera.mini.android.Browser}" drawable="opera_mini" />
```

Foggalong:
@bil-elmoussaoui How much more complicated would 1 be?

bilelmoussaoui:
Never mind what i said, it was late and I was tired :P I have a better solution, i will create a PR with the fixes

PR #4065, opened 2017-12-25
Add icon cache test for linux.

bilelmoussaoui:
Recently, one of the merged commits caused an error with icon cache. It doesn't work if the icon contains a space on it. 

In order to avoid this in the future, this test should make sure that the data.json file is correct.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,4 @@ script:
     - python3 ./tests/db_entry.py
     - python3 ./tests/icons_entry.py
     - python3 ./tests/ordered_db.py
+    - python3 ./tests/icon_cache.py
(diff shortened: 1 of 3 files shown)


bilelmoussaoui:
The tests fails because of this 
```
924C_StarCraft II.0 contains an empty space on it
```
I don't know what should I do, rename the icon without an empty space or remove the link?

Foggalong:
Remove the symlink since it can only work with that specific icon name. I believe that's the format for Wine software too so not essential to keep :)

bilelmoussaoui:
Done :+1:

Foggalong (approved):

Issue #3704, opened 2017-04-29
Icon Requests: Android Apps


mueller-ma:
| Application name | Android Package | Original icon | Comments/Icons that can be reused |
| --- | :-- | :-- | :-- |
| `1010! Klooni` | `io.github.lonamiwebs.klooni`| ![io_github_lonamiwebs_klooni_io_github_lonamiwebs_klooni_androidlauncher](https://cloud.githubusercontent.com/assets/22525368/25559270/5272df22-2d37-11e7-87f1-9004199ae9c0.png) | |
| `aCalendar+` | `org.withouthat.acalendarplus`| ![org_withouthat_acalendarplus_org_withouthat_acalendar_acalendar](https://cloud.githubusercontent.com/assets/22525368/25559289/52c6fa9e-2d37-11e7-9754-805318a630b1.png) | |
| `aCalendar+ Aufgaben` | `org.withouthat.acalendarplus`| ![org_withouthat_acalendarplus_org_withouthat_acalendar_tasks_tasksactivity](https://cloud.githubusercontent.com/assets/22525368/25559290/52c7657e-2d37-11e7-8a8e-258bbd5d0854.png) | |
| `Android Wear` | `com.google.android.wearable.app`| ![com_google_android_wearable_app_com_google_android_clockwork_companion_launcher_launcheractivity](https://cloud.githubusercontent.com/assets/22525368/25559244/52123064-2d37-11e7-9281-749bc741fcc0.png) | |
| `Anecdote` | `io.gresse.hugo.anecdote`| ![io_gresse_hugo_anecdote_io_gresse_hugo_anecdote_mainactivity](https://cloud.githubusercontent.com/assets/22525368/25559269/5271f436-2d37-11e7-851a-04293ed682b1.png) | |
| `Apk Extractor` | `axp.tool.apkextractor`| ![axp_tool_apkextractor_axp_tool_apkextractor_mainactivity](https://cloud.githubusercontent.com/assets/22525368/25559230/51dc078c-2d37-11e7-928d-d84d360309f9.png) | |
| `Ask me anything meaningful` | `de.lsubel.amam`| ![de_lsubel_amam_de_lsubel_amam_mainactivity](https://cloud.githubusercontent.com/assets/22525368/25559264/525c5b44-2d37-11e7-9cf0-b0b9ad5357dc.png) | |
[...]

PR #4079, opened 2017-12-28
Android entry for DuckDuckGo

mrmeszaros:
Also, did some cleanup around the icon.
Mentioned in #3704


palob (commented):
Now that you are at it, this grey path could do with much less nodes.
bildschirmfoto vom 2017-12-28 16-07-36

mrmeszaros:
@palob Just noticed, that the symbol uses pure white:

| Current: `#ffffff` | Opt A: `#eeeeee` |
| --- | --- |
|duckduckgo circle 192|duckduckgo circle 192 a|

palob:
Opt A 👍

palob (approved):

palob:
Is data.json supposed to be that way with an empty line before the Android entry?

mrmeszaros:
What empty line?
Are talking about the android entry list? I think that should be one string per line.

palob:
You're right it only appears to be blank due to how GitHub displays it.

icons/square/48/duckduckgo.svg
```diff
@@ -0,0 +1,28 @@
+<svg version="1.1" viewBox="0 0 48 48" xmlns="http://www.w3.org/2000/svg">
```
bilelmoussaoui, line 1: Can you resave the file with using 2 spaces as indentation? Like all other files :+1:

mrmeszaros: @bil-elmoussaoui other files use **1 space** indentation - for consistency I can use 1 space.
Anyway, I started using 1 tab because I can set my editor tab length to any value (which is pretty convenient).

**Edit**:
I did some investigation: some of them uses 2 space (eg. _0ad_), some 1 space (eg. the templates) - and because of me, now some use 1 tab (eg. _starcraft-2_)

mrmeszaros: Also, here is a solution to having different indentation on file system and git repository:
https://stackoverflow.com/questions/2316677/can-git-automatically-switch-between-spaces-and-tabs#2318063
I can add a guide on how to use it.
But this needs unified indentation.

palob: This can be set in the Inkscape (Scour) optimisation presets as well.

Foggalong (approved):

PR #4120, opened 2017-12-30
Clean up the 48px templates

mrmeszaros:
* shorter id for the background linear gradient
* remove width/height from the `svg` tag in square


W-Floyd:
The changes stay when optimizing using our settings, so I see no problem with this.

palob (approved):

PR #4141, opened 2018-01-06
Replace calendar icons on updated template.

W-Floyd:
Minor really, but I have the script set up, so I thought I'd get these updated.


palob (approved):

palob:
If anybody feels like doing so they could clean up + align the numbers some day.
bildschirmfoto vom 2018-01-06 11-18-41

W-Floyd:
Yeah, the numbers are just as copied from who knows when. I would need to rework my scripts to rebuild things, but it'd be doable.

PR #4153, opened 2018-01-11
Improved BFB efficiency

Foggalong:
Slight change to the workflow which has a minor increase in terms of computation time which only gets better the more icon sizes we're supporting. It also has the benefit of looking a lot nicer and being consistent with the code elsewhere :smile:

2018-01-11-155559_1068x332_scrot

### DIFF
--- a/gen.py
+++ b/gen.py
@@ -141,8 +141,8 @@ elif platform == "linux":
             if path.exists(source):
                 if icon["linux"].get("bfb") and (size == "48"):
                     output_bfb = "{0}.png".format(icon["linux"].get("bfb"))
-                    convert_svg2png(source, output+output_bfb, 144, 144)
-                copy2(source, output+root)
+                    convert_svg2png(source, output + output_bfb, 144, 144)
+                copy2(source, output + root)
                 for link in icon["linux"].get("symlinks", []):
                     output_symlink = "{0}{1}.svg".format(output, link)
                     try:
--- a/gen.py
+++ b/gen.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 """
-Copyright (C) 2017
+Copyright (C) 2018
 This program is free software: you can redistribute it and/or modify
 it under the terms of the GNU General Public License (version 3+) as
 published by the Free Software Foundation. You should have received
--- a/gen.py
+++ b/gen.py
@@ -1,11 +1,12 @@
 #!/usr/bin/env python3
+
 """
-# Copyright (C) 2016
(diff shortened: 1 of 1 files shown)


gen.py
```diff
@@ -136,19 +137,14 @@ def convert_svg2png(infile, outfile, w, h):
...
-                if bfb_icon:
-                    output_bfb = "{0}/{1}/apps/{2}.png".format(linux_dir,
-                                                               size,
-                                                               bfb_icon)
-                    if int(size) == 48:
-                        convert_svg2png(source, output_bfb, 144, 144)
-                copy2(source, output)
+                if icon["linux"].get("bfb") and (size == "48"):
+                    output_bfb = "{0}.png".format(icon["linux"].get("bfb"))
+                    convert_svg2png(source, output+output_bfb, 144, 144)
```
bilelmoussaoui: There's a missing space between the `+`

gen.py
```diff
@@ -136,19 +137,14 @@ def convert_svg2png(infile, outfile, w, h):
...
+                copy2(source, output+root)
```
bilelmoussaoui: Same here

bilelmoussaoui (dismissed):

gen.py
bilelmoussaoui: There's no need for parenthesis in `size == "48"`

gen.py
Foggalong: I was trying to work out why neither my linter nor the GH check had picked that up. Turns out spaces around arithmetic operators [are no longer required](https://hg.python.org/peps/rev/37af28ad2972) in PEP8. However assignments, comparisons, boolean, and curiously `%` still need them though.

Is there a way we can tell the checker here to also check the disabled by default E226?

gen.py
Foggalong: My thinking was that
```python
# Option 1
if gbfb and (size == "48"):
    print("Do the thing")
print("Do the next thing")
```
is clearer than just
```python
# Option 2
if gbfb and size == "48":
    print("Do the thing")
print("Do the next thing")
```
because it makes the two sides of the `and` clear. This also works because logic:
```python
# Option 3
if (gbfb and size) == "48":
    print("Do the thing")
print("Do the next thing")
```

gen.py
bilelmoussaoui: There must be a way, I will take a look at this later :+1:

Foggalong:
Using BDFL status to merge this in because I'd like to work on some other stuff with the code, but happy to come back to this @bil-elmoussaoui if you had other issues which weren't brought up and aren't addressed in the correction commit.

PR #4155, opened 2018-01-12
add nemo symlink

bilelmoussaoui:
Nemo icon name was renamed recently to use nemo instead of folder


Foggalong (changes requested):
[Icon's already there](https://github.com/numixproject/numix-core/blob/master/data.json#L10304-L10308). There should probably be a check added for conflicting data.json entries.

bilelmoussaoui:
Will add that after my finals!

PR #4161, opened 2018-01-19
Update Icon for Iridium Browser

cg-alves:
| --- | :-- | :-- |
|  original |  circle | square

Update to better reflect the browser's current logo.


palob (changes requested):
Hi, thanks for contributing.

+ The symbol is circular-ish. According to our guidelines this means the symbol shoul fit into a 30px by 30px circle. The lock and the blue ball may exceed the circle.

+ Many elements need streamlining. How did you create the symbol. By vectorisation? The elements need to be redone from scratch in order to get simpler shapes and reduce the number of nodes.

+ Interface elements need to be aligned with the pixel grid (``#`` key in Inkscape) as much as possible.

+ The symbol shadow should be black with 10% opacity and be displaced 1px down (Square) and 1px down + 1px right (Circle).

+ Please save the file as *Optimised SVG* in Inkscape with the CSS to XML option unchecked

cg-alves:
Hi, palob. Thank you for the feedback. I've committed the changes you asked for. Is everything as required?

palob (changes requested):
There are still alignment issues. A few examples:

+ In the current design the blue circle should be exactly 5px by 5px and placed at integer coordinates

+ The outer circular shape of the symbol parts should exactly follow the 30px x 30px template circle (shown red in the screenshot above).
bildschirmfoto vom 2018-01-21 13-33-11 The black lines should then be exactly 3px wide and be aligned with the grid as I did it with the one yellow path as an example.
+ See the yellow exemplary draft for the keyhole

mrmeszaros:
@5had0w5talk3r Also, after the icon visuals have been approved, please do some source cleanup:
* make the background a circle instead of a path
* clean up the gradient definition: remove `gradientTransform`
* optionally rename the gradient to `bg`
* remove unnecessary `<metadata>` tag
* clean up the `<svg>` attributes

For help, see our templates: `templates/circle/48.svg`, `templates/square/48.svg`

palob (changes requested):
The lock edge should be like this (maybe the upper one too?)
bildschirmfoto vom 2018-06-21 01-52-31

and the Circle baseplate is a bit dislocated.

Foggalong:
I'm not sure what you mean by like this - as in slightly curved?

palob:
Just not slightly overrunning the pixel border.

Foggalong:
Ah got it, will look at that again tomorrow :+1:

Foggalong:
Something was bugging me about this icon and I only just realised that it was because the gaps between the shield and the rest of the symbol didn't actually correspond to the current state of the icon. Aligned the shield points and recut all the gaps to reflect those changes and previous :+1:

palob:
Shield is still not aligned spot-on. Would be require slight bending and departure from the original design.

Foggalong:
Ah, in which case I think I misunderstood what you were talking about with the shield thinking you meant the points rather than the edges (which is obvious now I reread what you said).

Are you thinking rotate the shield so it can be straightened out fully?

palob:
Yep, as to minimalise the number of half-half coloured pixels.

Foggalong:
Sounds like a good idea, I'll have another crack at it over the weekend

Foggalong:
@palob Thoughts on these? Tried using a more orthographic perspective with both a square style lock and one more akin to the original design. Welcome to being told to go back to the drawing board, I kinda like it but not sure how close it is to what you were suggesting.

2019-02-24-180141_691x365_scrot

Foggalong:
I think the one on the right could be improved a little by tweaking the height to be 1 pixel higher, but don't have time to mock-up a preview right now.

palob:
That's rad, much better than the original design. I think I prefer the left one where the padlock curvature  doesn't interfere with the circle.

mrmeszaros:
@Foggalong I think I like the one on the left a bit more:
* the way the padlock interacts with the black pattern looks better
* there is no leftover black next to the padlock's gray band.

Because of this the left one looks a bit more busy

Maybe if the body of the lock would be 1px higher as You said, it might sway me back.

palob:
Haha, great minds think alike :laughing:

Foggalong:
@palob Pushed those changes!

palob (approved):

PR #4164, opened 2018-01-20
Fixes for most broken Square icons pt4

palob:
For #3482 + Circle fixes
square8


mrmeszaros:
Hey @palob good work!

Could You update the symbol base plates with the current templates (no `gradientTransform`, custom `id` for the gradient, as well as `circle` instead of `path`)?

palob:
I think we should keep this separate and do this later script-wise to keep the way we do this uniform.

Foggalong (approved):

PR #4171, opened 2018-01-21
Add icon for Firefox Focus

mrmeszaros:
| --- | :-- | :-- |
|org_mozilla_focus_org_mozilla_focus_activity_mainactivity|firefox-focus circle 192|firefox-focus square 192|

Colors and symbol path borrowed from the official Mozilla website:
https://www.mozilla.org/en-US/firefox/mobile/


palob:
What do you mean by "borrowed"?
The question now is if we should redesign all our Firefox icons to be closer to the new Quantum branding.

mrmeszaros:
Sorry for the vague language:
* **"borrowed"** means I copied the path data from the svg logo in the nav-bar
  * I think this means a symbol rework is needed (the logo is mozilla's property, I guess???)
  * although not so sure, about legal issues, but it may be a bit too busy with the extra bits
* **Quantum**: I like the current Firefox logo, but I think a redesign would be lovely

mrmeszaros:
I did some cleanup on circle, but left square as-is:
firefox-focus 384-192-96-48-24

palob:
Now that the outer edge is a Circle almost, why not go full Circle, do a bit cleaning up on the head and use this as a base for a Quantum redesign of all our Firefox icons?

I envisage the redesign as - compared to the current Numix icons - removing the continents paths, having only one path for the fox (no seperate paths for paw, nose) and a fringe-less tail. And of course a recolour. For the base Firefox icon a more orangy red and a beefier blue.

mrmeszaros:
* The outer part is a circle - from ear to _tail feather_
  _(Edit: **You were right** - just noticed that I gave wrong coords for my intersection tool)_
* Did not try it without the tail feather, but good idea
  * I left the _inside-feather_ intentionally, to break up the symbol a bit
#### Next on:
- [ ] clean up the hairs everywhere
- [ ] reduce node count
- [ ] smoothen the curves, where necessary
- [ ] adjust to grid where possible

Foggalong (changes requested):
I'd like to see some sort of lighter orb in the middle to match up with the other icons.

palob:
Hairs should be removed from the cheek as well.

palob:
It's alright for me design-wise now.
Design is different from our existing Firefox icon but maybe Firefox *Focus* justifies a less detailed design.

palob (approved):

mrmeszaros:
@palob I would like to lower the node count on the paw / neck - but with that, I feel that the icon is finished.

mrmeszaros:
Here is the current (final?) look:
firefox-focus 384-192-96-48-24

palob (approved):

Foggalong (approved):

PR #4180, opened 2018-01-31
small fix for LibreOffice Base

palob:
Before
lo-base-before

After
lo-base-bafter

Same for Square


Foggalong:
I literally use this software and have never noticed that 😆

Foggalong (approved):

PR #4189, opened 2018-02-03
Icons for Cockatrice

palob:
https://github.com/Cockatrice/Cockatrice

index svg

cockatrice-ci
cockatrice-sq


Foggalong (changes requested):
Realise that this is just the upstream design, but not merging because it's especially awful. Possibly some iconography involving servers or cards (or both) instead?

palob:
cocka

Foggalong:
@palob That's good with me :)

Foggalong (approved):

PR #4200, opened 2018-02-11
Add icon for Ziggurat

mrmeszaros:
| --- | --- | --- |
|ziggurat original 192|ziggurat circle 192|ziggurat square 192|


palob (changes requested):
Design is cool but I think the symbol could be a tad brighter / more contrasting.

mrmeszaros:
Here they are with 3% brighter symbol colors:
ziggurat 192

@palob may I include utouch and shine versions?

palob (approved):

palob:
You are free to port/update as much icons to utouch/shine as you wish :)

Tbh I haven't had a glimpse on those in a ling time. Don't even recall how they look like.

mrmeszaros:
It seems to me that shine is usually a 2-colored base divided in 2, and a usually whitish symbol on top.

BTW: it seems to me that ImageMagick, that I use to create these bundled images distorts the colors (eg, the background on the image above is much more saturated than in the SVG)

PR #4204, opened 2018-02-12
logos of pr0gramm.com and 4chan.org

0b11001111:
| --- | :-- | :-- |


palob (commented):
What's going on here? We've already merged those icons and you've already merged commits in here?

PR #4256, opened 2018-03-04
More Square maintenance

palob:
For #3482 + Circle fixes

square12
some-more


Foggalong (changes requested):
* shadow on the circle icon is for square
* the smile for reddit looks creepy - make it flatter
* general comment: why the change to the wine colour?

palob:
Regarding the wine colour I thought I'd make it a bit more [Windowsy](https://de.wikipedia.org/wiki/Microsoft#/media/File:Microsoft_logo_(2012).svg ). Got a bit exaggerated maybe.

Foggalong:
Comparison of the old colours, new colours, and the ones from the official logo:

2018-03-13-145319_1093x484_scrot

It might actually be the green which needs to be a shade darker rather than the yellow

palob:
Everything's fixed!

Foggalong (approved):

PR #4326, opened 2018-04-22
Entry and icons for Ksniffer.

W-Floyd:
#2294

iconified_ksniffer

Derived from `preferences-system-search`


palob (changes requested):
The Square icon has some clip stuff in it. Just deleting it from the code should do it.

W-Floyd:
I should probably double check `preferences-system-search` then as well.

W-Floyd:
See #4332 for fixes to `preferences-system-search` as well.

palob (approved):

PR #4372, opened 2018-06-08
Icons for Friture

palob:
http://friture.org/

friture
friture-ci
friture-sq


Foggalong (changes requested):
The colours on this seem a bit off to me. The orange and blue don't go together well without the grey buffer in the original icon (though that's ofc not suitable for inclusion in a numix icon)

palob:
Changed to this:
friture2

Foggalong (approved):

PR #4394, opened 2018-06-17
Icons for Puzzle Master

palob:
https://github.com/Venemo/puzzle-master
puzzle-master
puzzle-master-ci
puzzle-master-sq


Foggalong (approved):
We use the puzzle piece iconography for extensions so it feels weird using it for puzzle games also even though the latter makes much more sense. This is possibly something we'll have to come back to later.

PR #4419, opened 2018-06-21
Icons for SigViewer

palob:
https://github.com/cbrnr/sigviewer
sigviewer128
sigviewer128-ci

sigviewer128-sq


Foggalong (changes requested):
What does this like with the same design but the maximum amplitude of the wave being smaller (so vertically squashed)? I like it but it feels a bit big

palob:
Now looks like this
sigviewer128-ci
sigviewer128-sq

Foggalong (approved):

PR #4421, opened 2018-06-21
bygfoot positioning fix

palob:


Foggalong (changes requested):
With the original icon being different we should just straighten the numbers and shirt out fully

palob:
Same build failure here.

Foggalong (approved):

PR #4429, opened 2018-06-27
Icons for welle.io

palob:
https://github.com/AlbrechtL/welle.io

welle-io
welle-io-ci
welle-io-sq


palob:
Looks like ``thonny``  and ``threedepict`` need to be swapped indeed.

Foggalong:
@palob I'm gonna merge the open pulls and then just do a db correction before building :+1:

Foggalong (approved):

PR #4436, opened 2018-07-01
Icons for FreeDV

palob:
https://freedv.org/
freedv
freedv-ci
freedv-sq


Foggalong (changes requested):
I appreciate what you're trying to do with this icon but the design doesn't work. A symbol relating to radio, recording, or voice would work better. Maybe a radio with the current symbol design on it?

palob:
freedv-ci
freedv-sq

Foggalong (approved):

PR #4444, opened 2018-07-04
Kitty: Narrow terminal for real

palob:
I made a botched job of the cruft removal such that somehow the old design slipped in.

This one has the narrow terminal and cruft removed.


Foggalong (approved):
Thanks for catching that, improper review job on my part

PR #4459, opened 2018-07-11
Icons for Calls

palob:
https://source.puri.sm/Librem5/calls/tree/master

calls
calls-ci
calls-sq


mrmeszaros:
There is a similar icon for the android phone activity #4123

palob:
Yeah, like the yate icon but it not as "striaght".

Foggalong (approved):

PR #4461, opened 2018-07-11
Fixup Terraria icons

mrmeszaros:
* Fix shine-through pixel lines
* Apply optimizations


palob (commented):
If you are at optimisation, the tree shadow could use a cleanup. There's path boundaries inside it. A path union should do it.

palob (approved):

PR #4468, opened 2018-07-13
code.svg were added both for circle and squre for MS VScode

osiixy:
code-circle
code-square


bilelmoussaoui:
There's no reason to add `code.svg`. There's already an icon for VSCode.

osiixy:
What I see is a default editor icon and not a specific one for VScode which can be easily mistaken with Gedit.

palob:
We've already got this icon as ``visualstudiocode.svg``. So far we didn't add a symlink for ``code`` because of the generic nature of this icon name.

Issue #3938, opened 2017-10-06
Update Visual Studio Code Icon


jose-a-sa:
| Application name | Icon name | Desktop file name | Original icon |
| --- | :-- | :-- | :-- |
| `Visual Studio Code`| `code` | `code.desktop` | ![original](https://code.visualstudio.com/images/1_17_mac-stable-orange.png) |

Description: `Code editing. Redefined.`

Website: `https://code.visualstudio.com/updates/v1_17`

Foggalong:
See discussion in #3902 or #2964 depending what the intent of your report was.

PR #4469, opened 2018-07-13
code is also used to start MS VScode eg. under Ubuntu 16.04, so symli…

osiixy:
code is also used to start MS VScode eg. under Ubuntu 16.04, so symlink "code" has been removed from "text-editor" in favor of "visualstudiocode"

### DIFF
--- a/data.json
+++ b/data.json
@@ -15992,7 +15992,6 @@
                 "1E64_notepad.0",
                 "accessories-text-editor",
                 "blogilo",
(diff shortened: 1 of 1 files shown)


osiixy:
Not in order

palob:
Please see the reasoning in #3938.

osiixy:
Thanks for the link, now I understand. Sorry for bothering you!

Issue #2964, opened 2016-04-14
Icon for Visual Studio Code (vscode)


edgard:
vS code has been release today as 1.0 and the icon is not being used for the deb package. I noticed that there is already a visualstudiocode.svg, but the .desktop file is as follows:

/usr/share/applications/code.desktop
```
[Desktop Entry]
Name=Visual Studio Code
Comment=Code Editing. Redefined.
GenericName=Text Editor
Exec=/usr/bin/code %U
Icon=code
Type=Application
StartupNotify=true
Categories=Utility;TextEditor;Development;IDE;
MimeType=text/plain;
```

Suppose it needs a new symlink?

palob:
Hmm, just ``code`` is quite the generic icon name.

edgard:
heh, yeah. but maybe because it's so generic, it seems like nobody dared to use before. :D

Foggalong:
@edgard That's not a good thing, that's a bad thing. Generic icon names are likely to cause conflicts. I'd wager just symlink this to the generic text editor icon. It's better than nothing and I'm not sure we can comfortably fix this as just a VSCode icon.

edgard:
While I also do not like its called 'code' (even the cli), it's how it is right now. They oficially packaged like that. :)

wa4557:
Is it possible to include a "fix" for that in the hardcode fixer script?

Foggalong:
@edgard Yeah, they might have packaged it like that but it doesn't mean they've done it right. Lots of devs use `icon.svg` as an icon name but it's a pain for the same reason. We can't give it a specific icon because they used too generic a name which others are likely to also use. Please report it to them and ask for a fix.

@wa4557 wecould, yeah

icasdri:
I'm on a different distribution (namely Arch), that is currently packaging (unofficially by community members) with a different `Icon` line. So maybe hardcode fix and add icon for something like this while still generically linking "code" to text editor icon?
<pre>
[Desktop Entry]
Categories=Development;
Exec=/usr/bin/visual-studio-code %f
<b>Icon=visual-studio-code</b>
Terminal=false
Type=Application
StartupWMClass=code
Name=Visual Studio Code
Comment=Editor for building and debugging modern web and cloud applications
Categories=Development;WebDevelopment;IDE;
StartupNotify=true
</pre>

ChuckDaniels87:
[...]

PR #4470, opened 2018-07-13
fix compatibility issue for MS VScode

osiixy:
code is also used to start MS VScode eg. under Ubuntu 16.04, so symlink "code" has been removed from "text-editor" in favor of "visualstudiocode

symlink "code" was not in the right place

### DIFF
--- a/data.json
+++ b/data.json
@@ -15992,7 +15992,6 @@
                 "1E64_notepad.0",
                 "accessories-text-editor",
                 "blogilo",
-                "code",
                 "gedit",
                 "gedit-icon",
                 "gedit-logo",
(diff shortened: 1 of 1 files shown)


osiixy:
in plain language: I was stunned that I got the same boring icon for VScode just as for Gedit, it wasn't nice to see and it was hard to differentiate them (I even created icons for VScode because I had thought they were missing)

Foggalong:
Again, please read the [reasoning here](https://github.com/numixproject/numix-core/issues/2964#issuecomment-210359848). We're not making `code` a symlink to VS Code because it's too generic an icon name and will likely cause symlinks elsewhere. This is an issue but if you want it changing please petition the VS Code devs to change their name to something specific to their app like `vscode`.

Issue #4047, opened 2017-12-23
Use docker with travis


bilelmoussaoui:
Currently, with each commit Travis runs the tests and installs all the dependencies all over again. Which takes a lot of time. In order to reduce this, I'm going to create a Docker container based on Ubuntu 14.04  and install all the dependencies on it. 

This way, whenever Travis has to run the tests, it can use the docker container. The time diff will be huge, as only tests will have to be done.  The container will be kept up to date by me. This more to reduce the time of tests and give to the reviewers an easy life before merging an icon, and to the contributors an easy and fast way to verify their PR

I can provide a PR this weekend improve this.

Foggalong:
Made redundant by #5753

Foggalong:
Made redundant by #5753

PR #4530, opened 2018-07-30
Small batch of Square fixes

palob:
A few more for #3482 + Circle fixes


palob:
What's up with Travis?

palob:
Need to add fixes for Evolution and Thunderbird anyways.

Foggalong:
Sometimes it just does have a hissy fit when it's setting up and throws an error. Part of the reason why #4047 needs resolving

Foggalong (approved):

PR #4543, opened 2018-08-06
Instagram redesign + symlink for Ramme

### LINKED ISSUE #3604: Icon Request - Ramme
Icon for [Ramme](https://github.com/terkelg/ramme).

```
[Desktop Entry]
Name=Ramme
Comment=Unofficial Instagram Desktop App
Exec="/opt/Ramme/ramme"
Terminal=false
Type=Application
Icon=ramme
```

palob:
Fixes #3604

insta-ci
insta-sq


Foggalong:
Would this work if it used a pink baseplate still? It feels weird having a white baseplate when the original icon is so colourful.

Foggalong (changes requested):
Would this work if it used a pink baseplate still? It feels weird having a white baseplate when the original icon is so colourful.

palob:
insta-ci
insta-sq

Foggalong (changes requested):
This actually feels more wrong unfortunately. If you don't have any ideas (I know I don't) we'll go with the white baseplate for just being the least bad option.

(Not a slight on your icons, just the unfortunate incompatibility)

palob:
Rolled back to white.

Foggalong (approved):

PR #4568, opened 2018-09-04
Symlinks for adriconf

palob:
https://github.com/jlHertel/adriconf
adriconf

### DIFF
--- a/data.json
+++ b/data.json
@@ -788,12 +788,14 @@
         "linux": {
             "root": "application-default-icon",
             "symlinks": [
+                "adriconf",
                 "applications-accessories",
                 "applications-other",
                 "cs-default-applications",
                 "gnome-desktop-item-edit",
                 "gnome-run",
                 "gnome-settings-defaulg-applications",
+                "org.x.adriconf",
                 "preferences-desktop-launch-feedback"
             ]
         }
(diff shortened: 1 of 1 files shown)


data.json
```diff
@@ -788,12 +788,14 @@
         "linux": {
             "root": "application-default-icon",
             "symlinks": [
+                "adriconf",
                 "applications-accessories",
                 "applications-other",
                 "cs-default-applications",
                 "gnome-desktop-item-edit",
                 "gnome-run",
                 "gnome-settings-defaulg-applications",
```
Foggalong: I think this might be a typo, meaning to say `gnome-settings-default-applications`

data.json
```diff
@@ -788,12 +788,14 @@
         "linux": {
             "root": "application-default-icon",
             "symlinks": [
+                "adriconf",
```
Foggalong: I don't think we should use `application-default-icon` as a root icon for an actual application, has the potential to be misunderstood as a placeholder. Given that it's about graphics drivers the jockey icons might be more appropriate.

Foggalong (changes requested):

palob:
I'm doing a dedicated one for adriconf

Foggalong (approved):

PR #4571, opened 2018-09-07
Dedicated icons for adriconf

palob:
In lieu of the symlinks in #4568

adriconf-ci
adriconf-sq


Foggalong (changes requested):
Can we tried this with an actual cog symbol rather than a cut out?

palob:
Here's a white cog
adriconf-test

Foggalong:
@palob That's good with me :+1:

palob:
Pushed :+1:

Foggalong (approved):

PR #4574, opened 2018-09-10
Icons for remote-touchpad

palob:
https://github.com/Unrud/remote-touchpad

remote-touchpad
remote-touchpad-ci
remote-touchpad-sq


Foggalong (changes requested):
This design feel s weird to me and I think it might be the mouse. Do you think it would work with just the target square?

palob:
I don't think so. The original icon doesn't really tell what the app is all about.

Maybe something based on the libinput-gestures icon could work.

Foggalong:
@palob Yeah I think that might work better than the cursor

palob:
Hmm, didn't work out yet.
Maybe just symlinks to ``touchpad-indicator`` or ``input-touchpad``?

Foggalong:
@palob Yeah if that's not working a link to input-touchpad will suffice.

Foggalong (approved):

PR #4577, opened 2018-09-15
Icons for movie-monad

palob:
https://github.com/lettier/movie-monad

moviemonad
moviemonad-ci
moviemonad-sq


Foggalong:
Really good solution to not using symbol gradients! Kudos :heart:

Foggalong (approved):

PR #4591, opened 2018-10-01
Icons for electrum-ltc

palob:
https://github.com/pooler/electrum-ltc

electrum-ltc
electrum-ltc-ci
electrum-ltc-sq


Foggalong (changes requested):
I think this one can definitely manage without a gradient, just on the orbits use a flat shade which is slightly darker than the one used for the L

palob:
Alright, will change this for the original electrum icons as well then.

palob:
The icons for katomic, Atomic Game Engine and Scientific Linux are based on this design as well IIRC.

palob:
Removed the gradients, think it's all.
One example:
electrum--no-gradient

Foggalong (approved):

PR #4592, opened 2018-10-01
Icons for fondo

palob:
https://github.com/calo001/fondo

fondo
fondo-ci
fondo-sq


Foggalong:
Could we manage it without the gradient? Doesn't feel especially numixy as is

Foggalong (changes requested):
Could we manage it without the gradient? Doesn't feel especially numix-y as is

palob:
fondo-ci
fondo-sq

Foggalong (approved):

PR #4603, opened 2018-10-08
teamsql icon closes #4647

Gabb1995:
| Original icon | Circle icon |
| --- | :-- |
|  original |  circle |

Commit message is wrong, should be #4547


palob:
Why did you close this?

PR #4607, opened 2018-10-13
More Square fixes

palob:
For #3482 + Circle fixes


icons/square/48/gnome-usage.svg
```diff
@@ -1,26 +1,21 @@
-<svg width="48" height="48" version="1.1" viewBox="0 0 48 48.000001" xmlns="http://www.w3.org/2000/svg">
+<svg width="48" height="48" version="1.1" viewBox="0 0 48 48" xmlns="http://www.w3.org/2000/svg">
```
Foggalong: This `<svg>` tag is weird compared to the other icons. Not sure why width and height attributed are here?

Foggalong (changes requested):
The clipping on accessories-dictionary has come out a bit wrong. I think there needs to be a slight overlap between the two else you end up with parts of the red showing through.

There's also a weird issue with the viewbox on gnome-usage (square only)

Foggalong (approved):

PR #4608, opened 2018-10-13
Icons for 4DIAC IDE

palob:
https://www.eclipse.org/4diac/index.php

4diac
4diac-ci
4diac-sq


Foggalong (changes requested):
I realise it's in the original icon, but the dark blue being up against the orange of the 4 doesn't look quite right to me. How does it look without it or with that removed? It might be that the current design is still better, not sure

palob:
Removed it.
4diac-ci
4diac-sq

Foggalong:
Much better, thanks! :)

Foggalong (approved):

PR #4609, opened 2018-10-13
Flatpak symlinks

palob:

### DIFF
--- a/data.json
+++ b/data.json
@@ -5754,7 +5754,10 @@
     },
     "flameshot": {
         "linux": {
-            "root": "flameshot"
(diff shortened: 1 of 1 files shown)


data.json
```diff
@@ -16119,6 +16166,8 @@
         "linux": {
             "root": "supertux",
             "symlinks": [
+                "org.supertux.SuperTux",
+                "org.supertux.SuperTux-Milestone1",
```
Foggalong, line 16170: why in gods name did an icon like that make it into a SuperTux release???

Foggalong (approved):

PR #4610, opened 2018-10-14
Yet another bunch of optimisation for Square & Circle

palob:
For #3482 + Circle fixes


icons/square/48/gimagereader.svg
```diff
@@ -1,6 +1,6 @@
-<svg width="48" height="48" version="1.1" viewBox="0 0 48 48.000001" xmlns="http://www.w3.org/2000/svg">
+<svg width="48" height="48" version="1.1" viewBox="0 0 48 48" xmlns="http://www.w3.org/2000/svg">
```
Foggalong: This has that weird header with width and heigh attributes again. Maybe not something we need to worry about for individual icons, but we can replace them all en mass with one which we know works.

Foggalong (changes requested):
Issue with the shadow cause by symbol transparency in guake, guake-prefs (circle & square)

The paintbrush in mypaint should have a shadow. Similarly the download icon in grip to be consistent with what's done elsewhere.

Foggalong (changes requested):
It's a small change, but for mypaint the paint itself doesn't want to be having a symbol shadow (because it's presumably on the paper) just the brush :)

Foggalong (approved):

PR #4619, opened 2018-10-23
Icons for FFADO Mixer

palob:
http://ffado.org/

hi64-apps-ffado
hi64-apps-ffado-ci
hi64-apps-ffado-sq


Foggalong (changes requested):
I'm not familiar with the app, but looking at the icon I think the shape of the prongs might be to do with clipping to the border rather than the shape of the symbol. Is there an example of the icon on another base for comparison?

palob:
Couldn't find anything. Logo and original icon look the same.

palob:
Ah, you think I should go for pointed prongs all the way?

palob:
Like this or squares/rectangles in all directions?

ffado-mixer svg

Foggalong:
@palob I'd be alright with that, looks more fitting :)

palob:
Pushed these

ffado-mixer svg
ffado-mixer svg sq

Foggalong (approved):

PR #4646, opened 2018-12-08
fix the ubuntu vscode icon display

rikonaka:
| --- | :-- | :-- |

Here, when I using the numix-icon-theme both in ubuntu 16.04 and ubuntu 18.04 I found the vscode is using the code.svg instead of the vscode.svg, and I know there have an application is using the code.svg, but I think it's not as popular as the vscode, so I change the data.json make the code.svg generate from visualstudiocode.svg instead of the text-editor.svg.

Here is the file vscode.desktop details in ubuntu:
```
[Desktop Entry]
Name=Visual Studio Code
Comment=Code Editing. Redefined.
GenericName=Text Editor
Exec=/usr/share/code/code --unity-launch %F
Icon=code
Type=Application
StartupNotify=true
StartupWMClass=Code
Categories=Utility;TextEditor;Development;IDE;
MimeType=text/plain;inode/directory;
Actions=new-empty-window;
Keywords=vscode;

X-Desktop-File-Install-Version=0.23

[Desktop Action new-empty-window]
Name=New Empty Window
Exec=/usr/share/code/code --new-window %F
Icon=code
```
THX~

### DIFF
--- a/data.json
+++ b/data.json
@@ -16800,7 +16800,6 @@
                 "accessories-text-editor",
                 "blogilo",
                 "bottles_notepad",
-                "code",
                 "gedit",
                 "gedit-icon",
                 "gedit-logo",
@@ -17848,7 +17847,8 @@
                 "vsc",
                 "vscode",
                 "vscodium",
-                "vso"
+                "vso",
+                "code"
             ]
         }
     },


bilelmoussaoui:
Nah. VSCode guys should fix that upstream and follow the new freedesktop standards.

rikonaka:
@bilelmoussaoui So, do you mean by the downstream theme software to decide what the upstream software should do? I think it should be the numix to match other software. The numix theme software is rooted in other software. It should be adapted for these software instead of other softward to adapt the theme software.
Just like software should adapt hardware rather than hardware to adapt software. This is my opinion.

bilelmoussaoui:
Well, a software that doesn't follow the specs... We can't do anything about it. Its to upstream to fix their issue instead of making Linux users life harder. A lot of simple text editors use that as an icon name, we can't use the VSCode logo for those apps too.

rikonaka:
OK, I understand, I will try to go to vscode to make suggestions.

PR #4647, opened 2018-12-16
Icons for F-Droid Repomaker

palob:
https://gitlab.com/fdroid/repomaker

repomaker
repomaker-ci
repomaker-sq


Foggalong (changes requested):
That green of the head is too bright against that background. Try reducing the S or L values

palob:
Works better with a darker background anyway.
repomaker-ci
repomaker-sq

Foggalong (approved):

PR #4656, opened 2018-12-16
Icons for Fragments

palob:
https://gitlab.gnome.org/haecker-felix/Fragments
de haeckerfelix fragments
de haeckerfelix fragments-ci
de haeckerfelix fragments-sq


Foggalong:
Add a few extra pixels of width, maybe 1 or 2 in each direction, but keep the connection nubs the same size and shape. While the original jigsaw piece looks roughly square it's actually shorter on the out-nub side and longer on the in-nub side. This counteracts the illusion that it's a rectangle caused by having the two out-nubs.

2019-01-06-114625_703x747_scrot

Foggalong (changes requested):
Add a few extra pixels of width, maybe 1 or 2 in each direction, but keep the connection nubs the same size and shape. While the original jigsaw piece looks roughly square it's actually shorter on the out-nub side and longer on the in-nub side. This counteracts the illusion that it's a rectangle caused by having the two out-nubs.

2019-01-06-114625_703x747_scrot

palob:
de haeckerfelix fragments-ci
de haeckerfelix fragments-sq

Foggalong (approved):

PR #4671, opened 2018-12-30
Icons for Spine

sprocketc:
http://esotericsoftware.com/

| --- | :-- | :-- |
|  original |  circle | square|


palob:
Why did you remove the symlink?

palob (changes requested):
Thanks for your contribution. For the sake of sharpness it might help to fit the kidney things into boxes of integer pixel size.
bildschirmfoto vom 2018-12-31 09-41-06

palob (commented):
The lowermost and uppermost are fine now, the ones inbetween need to be tweaked a bit.

palob (commented):
Thanks, you tweaks did it. Can you please move symbol+shadow 1px to the left for both icons and 1px up for Square? Looks a bit off-centre right now.

palob (approved):

Foggalong (changes requested):
Could you reduce the saturation of the red ever so slightly? It's a bit too bright as is

Foggalong (changes requested):
Just noticed a shadow error with Square; the shadow for Square is just 1px under the symbol, rather than 1px right 1px down as with Circle.

sprocketc:
Are you sure? The position of the symbol should be x:18 y :12 and the shadow x:19 y:11

palob:
Huh, Square shadow ought to be jut 1px down :)

Foggalong:
If that's a change you've made it's not one you've pushed. The icons you've pushed have Circle's symbol at (18,11) with (19,10) shadow and Square's symbol at (18,12) with (19,11) shadow. Square's shadow should be at (18,11) instead, directly below the symbol rather than down and right.

sprocketc:
My bad. I thought the shadow guidelines for circle and square icons were the same.
Should be fixed now.

palob (approved):

Foggalong (approved):

PR #4677, opened 2019-01-02
Icons for GoldenCheetah

palob:
https://github.com/GoldenCheetah/GoldenCheetah

gc
gc-ci
gc-sq


Foggalong:
That's an ingenious way to approach that design, really well done!

Foggalong (approved):

PR #4693, opened 2019-01-06
Icons for PureRef

sprocketc:
https://www.pureref.com/

| --- | :-- | :-- |


sprocketc:
I added a symlink for .pur files.
Is there a template or a different project for file types? Should I remove the symlink?

palob:
I suspect what you are looking for with the file icon is a mimetype icon. Those are implemented in the base theme https://github.com/numixproject/numix-icon-theme

sprocketc:
Yes, that's what I mean. I removed the symlink.
Thank you.

palob (changes requested):
Thanks for coming back.
Just 2 small things:

+ the white parts should be ``f9f9f9ff`` instead of full white
+ the symbol shadows should be full black (0L) with 10% opacity

palob (approved):

Foggalong (approved):

PR #4706, opened 2019-01-12
Icons for SpriteIlluminator

sprocketc:
https://www.codeandweb.com/spriteilluminator

| --- | :-- | :-- |


palob (changes requested):
Symbol shape/perspective is a bit odd. The lower outer nodes of the "3D cube" should go up a bit.
Maybe the *S* needs to be tweaked a bit then to keep it parallel.

In this case you can transform all nodes into corners. There are some unneeded handles left.

sprocketc:
I decided to use an orthographic projection (despite the original) to make the symbol look as flat as possible but I forgot to change the upper corner. Should be fixed now.

Unneeded handles removed.
Thumbnails updated.

This is what I am trying to achieve:
rect1273

palob (commented):
Proportions are better now, but the *S* nodes still have got handles.

palob (approved):

Foggalong (approved):

PR #4720, opened 2019-01-22
Icons for Daty

palob:
https://gitlab.com/tallero/daty
daty
daty-ci
daty-sq


Foggalong (changes requested):
Like the idea, but extend one of the bars to narrow the 2px gap to 1px

palob:
daty-ci
daty-sq

Foggalong (approved):

PR #4737, opened 2019-01-29
Discreete Linux Icons

Foggalong:
This distribution forked the icon theme and added a load of packaging files, but also some icons which are used in their distribution. The icon files aren't stuff we'll want upstream but they can probably act as a base to aim to use. It'll also mean the patch they have to apply upstream is smaller.


palob:
I can't take a closer look right now but some icons seem to lack SVG-optimisation.

Foggalong:
@palob It's the same settings as the wiki in terms of settings - has there been an addition to those? And no worries re checking them over yet, gonna go update all of them so there's no rush yet.

palob:
The ``gnupg-xxx`` icons are longer than usual.

Foggalong:
@palob Oh right I get you! Those aren't my icons but the ones of the downstream distro. Will be going through them one by one and updating the designs

Foggalong:
Having looked at GnuPG icon online for reference, it turns out the icon we already have (added ~4 years ago) and thus the icon the additions are based on is based on old branding which was changed [over 10 years ago](https://web.archive.org/web/20070111125746/http://logo-contest.gnupg.org/). It must have just a theme which hadn't updated its icon yet which was used to choose colours for ours. Will update our design to match what they use now.

Foggalong:
This got properly messed up while resolving conflicts because of confusion with the two master branches. Going to close this PR and reopen in a new one.

PR #4741, opened 2019-01-29
Icons for Chores

### LINKED ISSUE #4082: Icon Request: Chores
| Application name | Icon name | Desktop file name | Original icon | Theme |
| --- | :-- | :-- | :-- | :-- |
| Chores | `com.github.georgekap1an.chores` | `com.github.georgekap1an.chores.desktop` | Chores| Numix-Circle |

This issue was automatically created using [IconRequests](https://github.com/bil-elmoussaoui/Icon-Requests).

really-need-an-api-key:
Fixes #4082.

| --- | :-- | :-- |
| original | circle | square |


palob:
The check looks a bit squeezed in. Maybe only 3 list items in total or a bigger gap to the lines on the right?

really-need-an-api-key:
Like this?
square48


square

palob:
How does it look like without the check touching the clip on top?

really-need-an-api-key:
Is having those two blue nodes at half pixel coords okay?

screenshot_2019-01-31_16-34-55

I think it looks more like a checkmark than when they're aligned perfectly, especially at higher res. Aligned looks like this:
square3

Foggalong (commented):
@really-need-an-api-key if you are doing the off grid edge, both sides need to match (i.e. both diagonal or both horizontal)

really-need-an-api-key:
What about this? I know the checkmark is misaligned but I think it's the sharpest I can get it and still kinda keep the general proportions. 
screenshot_2019-01-31_19-12-30

palob:
The last one looks the best since the left side looks like a "real" check now + the contrast is better. For non-vertical/-horizontal lines pixel-grid alignment is not as mush critical. There can be cases where nodes at non-integer positions avoid "begun", partially filled pixels and thus account for a sharper result.

really-need-an-api-key:
| | |
| :--- | :--- |
| circle | square |

palob (approved):

PR #4780, opened 2019-02-11
Script Cleanup

### LINKED ISSUE #3453: Icon Categories
_This issue was originally core-demo#5. It has been transferred here for archiving._

---

Foggalong, 9 Apr 2016

> Wondering if it might be advantageous to split apps based on the category they fall into? Like within each size have folders for `games`, `web-sapps`, `text-editors`, `browsers`, etc.

bil-elmoussaoui, 1 Aug 2016

> I think it's a great idea and we should add this to the database file asap. But will this work on all OS's or only linux?

### LINKED ISSUE #4047: Use docker with travis

### LINKED ISSUE #4779: Ordered DB Test Crash
I'm currently doing a refresh of some of the Python code and found that the [ordered_db.py](https://github.com/numixproject/numix-core/blob/63e2b9b4e3e5c94760800068c673c41dd3e58c67/tests/ordered_db.py) check breaks if the Android icons are out of order. This is because our current build script uses a structure with `/` characters in Android entries while the JSON loader seems to expect `,` characters for delimiters. The full crash is below.

```python3
Traceback (most recent call last):
  File "ordered_db_old.py", line 23, in <module>
    data = json.load(db_obj, object_pairs_hook=OrderedDict)
  File "/usr/lib/python3.6/json/__init__.py", line 299, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "/usr/lib/python3.6/json/__init__.py", line 367, in loads
    return cls(**kw).decode(s)
  File "/usr/lib/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.6/json/decoder.py", line 355, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 5769 column 13 (char 123776)
```

While the current work I'm doing on the tests does change this script it leaves this particular bug unfixed.

### LINKED ISSUE #4786: Duplicated Entry Test
Currently it doesn't look like we do Travis tests for duplicate entries. We probably should to avoid cases [such as this](https://github.com/numixproject/numix-core/pull/4783#issuecomment-464513399) in #4783.

Foggalong:
This PR fixes a couple of bugs with the script part of Core, increases efficiency, and generally cleans up the codebase. Lots of the code hasn't been touched since it was first written because it Just Works™, but for that reason some issues have gone unchecked.

I'll comment on the commits themselves with specific improvement details, but in short this is what this PR hopes to achieve:
* [x] Complete review of the test scripts
* [x] Complete review of the main generator code
* [x] Fixes #3453
* [x] Fixes #4047
* [x] Fixes #4779 
* [x] Fixes #4786
* [x] Add action for checking `.editorconfig` followed

### DIFF
--- a/gen.py
+++ b/gen.py
@@ -41,7 +41,8 @@ except FileNotFoundError:
 parser = ArgumentParser(prog="Numix-core")
 
 try:
-    theme, themes = "", listdir("icons")
+    # Exclude files and tag directories from the list of themes
+    themes = [dir for dir in listdir("icons") if "." not in dir]
 except FileNotFoundError:
     exit("No icons folder found. Please reclone and try again.")
 parser.add_argument("--theme", "-t",
--- a/gen.py
+++ b/gen.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 """
-Copyright (C) 2018
+Copyright (C) 2019 Numix Project
 This program is free software: you can redistribute it and/or modify
 it under the terms of the GNU General Public License (version 3+) as
 published by the Free Software Foundation. You should have received
@@ -18,33 +18,35 @@ from shutil import copy2
 from subprocess import PIPE, Popen, call
 try:
     from gi import require_version
-    require_version('Rsvg', '2.0')
+    require_version("Rsvg", "2.0")
     from gi.repository import Rsvg
     from cairo import ImageSurface, Context, FORMAT_ARGB32
     use_inkscape = False
 except (ImportError, AttributeError, ValueError):
-    ink_flag = call(['which', 'inkscape'], stdout=PIPE, stderr=PIPE)
-    if ink_flag == 0:
-        use_inkscape = True
-    else:
+    # Assigns true if Inkscape found, false otherwise
+    use_inkscape = not call(["which", "inkscape"], stdout=PIPE, stderr=PIPE)
+    if not use_inkscape:
         exit("Can't load cariosvg nor inkscape")
 
 
-parser = ArgumentParser(prog="Numix-core")
-
-# Importing JSON
+# Loading the JSON data file
 try:
-    with open('data.json') as data:
+    with open("data.json") as data:
         icons = load(data)
 except FileNotFoundError:
     exit("Please clone the whole repository and try again.")
 
+
+# Parsing build arguments
+parser = ArgumentParser(prog="Numix-core")
+
 try:
     theme, themes = "", listdir("icons")
-    parser.add_argument("--theme", "-t",
-                        help="Theme you want to build.", choices=themes)
 except FileNotFoundError:
     exit("No icons folder found. Please reclone and try again.")
+parser.add_argument("--theme", "-t",
+                    help="Theme you want to build.",
+                    choices=themes)
 
 platform, platforms = "", ["android", "linux", "osx"]
 parser.add_argument("--platform", "-p",
@@ -55,15 +57,17 @@ args = parser.parse_args()
 
 # User selects the theme
 if not args.theme:
-    exit("Please use --theme argument with "
-         "one of the following: {}".format(", ".join(themes)))
+    msg = "Please use --theme argument with one of the following: "
+    exit(msg + ", ".join(themes))
 else:
     theme = args.theme
+    # Each theme may support different sizes
+    sizes = listdir("icons/" + theme)
 
 # User selects the platform
 if not args.platform:
-    exit("Please use --platform argument with "
-         "one of the following: {}".format(", ".join(platforms)))
+    msg = "Please use --platform argument with one of the following: "
+    exit(msg + ", ".join(platforms))
 else:
     platform = args.platform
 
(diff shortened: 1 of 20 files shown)


tests/db_entry.py
```diff
@@ -1,34 +1,34 @@
 #!/usr/bin/env python3
+
 """
-# Copyright (C) 2016
-# This program is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License (version 3+) as
-# published by the Free Software Foundation. You should have received
-# a copy of the GNU General Public License along with this program.
-# If not, see <http://www.gnu.org/licenses/>.
+Copyright (C) 2019 Numix Project
```
Foggalong, line 4: Updated all copyright headers, removed unneeded `#` characters

tests/db_entry.py
```diff
@@ -1,34 +1,34 @@
...
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License (version 3+) as
+published by the Free Software Foundation. You should have received
+a copy of the GNU General Public License along with this program.
+If not, see <http://www.gnu.org/licenses/>.
 """
 
+from json import load
 from os import path
-import json
```
Foggalong, line 12: Alphabetised all (global) import lists

tests/db_entry.py
```diff
@@ -1,34 +1,34 @@
...
 
-from utils import error
+from utils import error, THEMES
```
Foggalong: The list of themes was used in multiple tests so made it part of `utils.py`

tests/db_entry.py
```diff
@@ -1,34 +1,34 @@
...
 
 ABS_PATH = path.dirname(path.abspath(__file__))
 DB_FILE = path.join(ABS_PATH, "../data.json")
-THEMES = ["circle", "square"]
 ICONS_DIR = path.join(ABS_PATH, "../icons/")
 
+with open(DB_FILE, 'r') as db_obj:
+    data = load(db_obj)
```
Foggalong, line 22: The actual test is no longer carried out inside the file-read `with` block.

tests/db_entry.py
```diff
@@ -1,34 +1,34 @@
...
-                error("The icon {} doesn't exist"
-                      " in the theme {}".format(
-                          entry, theme
-                      ))
-exit(int(has_errors))
+for entry in data:
+    for theme in THEMES:
+        icon = path.join(ICONS_DIR, theme, "48/{}.svg".format(entry))
+        if path.exists(icon):
+            continue
```
Foggalong, line 29: Using a `continue` rather than an extra level of indentation after a lone-`if` makes the code cleaner, especially later one when it's reaching 6-7 levels.

tests/db_entry.py
```diff
@@ -1,34 +1,34 @@
...
+
+        has_errors = True
+        error("'{}' doesn't have an icon in the {} theme".format(entry, theme))
+
+exit(has_errors)
```
Foggalong, line 37: `has_errors` is a bool value so it doesn't need converting to an integer before being passed to `exit()`

tests/icon_cache.py
```diff
@@ -1,41 +1,39 @@
...
 
-from utils import error
+from utils import error, space_check
+
 
 ABS_PATH = path.dirname(path.abspath(__file__))
 DB_FILE = path.join(ABS_PATH, "../data.json")
 
 with open(DB_FILE, 'r') as db_obj:
-    data = json.load(db_obj, object_pairs_hook=OrderedDict)
```
Foggalong, line 20: This test loaded the JSON file as an ordered dictionary which isn't needed here.

tests/icon_cache.py
```diff
@@ -1,41 +1,39 @@
...
+If not, see <http://www.gnu.org/licenses/>.
 """
+
+from json import load
 from os import path
-import json
-from collections import OrderedDict
 
-from utils import error
+from utils import error, space_check
```
Foggalong: `space_check` is now a `utils.py` function replacing `test_empty_space`

tests/db_entry.py
```diff
@@ -1,34 +1,34 @@
...
-                      ))
-exit(int(has_errors))
+for entry in data:
+    for theme in THEMES:
+        icon = path.join(ICONS_DIR, theme, "48/{}.svg".format(entry))
+        if path.exists(icon):
+            continue
+
+        has_errors = True
+        error("'{}' doesn't have an icon in the {} theme".format(entry, theme))
```
Foggalong, line 32: Added quotation marks `'...'` around DB entries in all cases, rather than speech marks `"..."` around only some cases.

tests/icons_entry.py
```diff
@@ -1,38 +1,38 @@
...
         icon_name = path.splitext(path.basename(icon))[0]
-        if icon_name not in entries and icon_name not in reported:
-            reported.append(icon_name)
-            has_errors = True
-            error("The icon {} doesn't have any "
-                  "entry in the database.".format(
-                      icon_name
-                  ))
-exit(int(has_errors))
+        if icon_name in entries or icon_name in reported:
```
Foggalong, line 31: As well as the previous comment about `continue` being cleaner than a whole new `if` level, this has an efficiency improvement. For those who know logic, you'll recognise that:

```
(¬A) ʌ (¬B) === ¬(A v B)
```

Since the continue handles the `NOT` on the right hand side, we're actually checking whether just `(A v B)`. Since `A` and `B` are just inclusion events (`x in X`) they're much quicker than evaluating the previous exclusion events (`x not in X`). This is further improved because Python halts logical evaluations as soon as it knows the statement as a whole is true, i.e. here if `icon_name in entries == True` it doesn't bother evaluating `icon_name in reported` and continues to the `continue`.

tests/ordered_db.py
```diff
@@ -1,57 +1,40 @@
...
 """
 
-from os import path
-import json
 from collections import OrderedDict
+from json import load
+from os import path
 
-from utils import sort, error
+from utils import error, sort, sort_check
```
Foggalong: The code for checking whether a list was sorted was included three times separately in this script rather than being turned into a function. It's now done as part of `utils.py`.

tests/ordered_db.py
```diff
@@ -1,57 +1,40 @@
...
-        error("Database entry {} not correctly "
-              "ordred".format(keys[i]))
-        print("Should be placed at {} instead "
-              "of {}".format(correct_position, i))
+    data = load(db_obj, object_pairs_hook=OrderedDict)
 
+has_errors = sort_check(list(data.keys()), "Database entry")
 
 for key, value in data.items():
+    if value.get("android"):
```
Foggalong: Moved Android to top to match JSON.

tests/ordered_db.py
```diff
@@ -1,57 +1,40 @@
...
-            if android_icons[i] != ordered_and_icons[i]:
-                has_errors = True
-                error("Android Icon \"{}\" not correctly "
-                      "ordered".format(android_icons[i]))
-exit(int(has_errors))
+        if not symlinks:
+            continue
+        has_errors = sort_check(symlinks, "Linux symlink", key)
+
+    if value.get("osx"):
```
Foggalong: Added OSX check.

tests/schema.json
```diff
@@ -4,6 +4,14 @@
...
             "type": "object",
             "properties": {
+                "android": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    },
+                    "minItems": 1,
+                    "uniqueItems": true
+                },
```
Foggalong, line 14: Moved Android to top to match `data.json`

tests/schema.json
```diff
@@ -26,7 +34,7 @@
                         "root"
                     ]
                 },
-                "android": {
+                "osx": {
```
Foggalong, line 37: Added OSX

tests/schema.py
```diff
@@ -1,23 +1,25 @@
...
 from jsonschema import validate, ValidationError
 from jsonschema.exceptions import SchemaError as SchemaError
+from os import path
 
 from utils import error, success
 
-DB_FILE = path.join(path.dirname(path.abspath(__file__)), "../data.json")
-SCHEMA_FILE = path.join(path.dirname(path.abspath(__file__)), 'schema.json')
+
+ABS_PATH = path.dirname(path.abspath(__file__))
```
Foggalong: Used twice so made a variable

tests/utils.py
```diff
@@ -1,18 +1,16 @@
...
+published by the Free Software Foundation. You should have received
+a copy of the GNU General Public License along with this program.
+If not, see <http://www.gnu.org/licenses/>.
 """
 
-
-def sort(icons_list):
-    """Sort list case insensitive."""
-    return sorted(icons_list,
-                  key=lambda icon_name: icon_name.lower())
```
Foggalong: Reordered `sort()` to be with `sort_check()`

tests/utils.py
```diff
@@ -23,3 +21,43 @@ def error(msg):
...
     print("\033[92m{}\033[0m".format(msg))
+
+
+def space_check(string):
+    """Checks if a string contains an empty space."""
+    if " " in string:
+        error("'{}' contains an empty space".format(string))
+        return True
+    else:
+        return False
```
Foggalong: Returns a bool rather than making use of global variables as previous.

tests/utils.py
Foggalong, line 15: Deleted `sort()` since was only used in `sort_check()`

tests/utils.py
```diff
@@ -23,3 +21,38 @@ def error(msg):
...
+
+
+def sort_check(list, stype, root=""):
+    """
+    Checks if a list of 'stype' strings is sorted, case insensitive. Takes as
+    optional string 'root' as a root value for use with sublists.
+    """
+    has_errors = False
+    reference = sorted(list, key=lambda item: item.lower())
```
Foggalong: The previous `sort` function absorbed into here.

tests/utils.py
```diff
@@ -23,3 +21,38 @@ def error(msg):
...
+
+        correct_index = reference.index(list[i])
+        has_errors = True
+
+        if root == "":
+            error_msg = "{} '{}' not correctly ordred"
+        else:
+            error_msg = "{} '{}' of '{}' not correctly ordred"
+
+        error(error_msg.format(stype, list[i], root))
```
Foggalong: Extraneous `format` arguments are simply ignores so using one `error` for both messages fine.

Foggalong:
I wouldn't normally thoroughly review my own code changes like this but I didn't actually write any of these tests, @bilelmoussaoui did. With that in mind I thought it only fair I explain what's being changed.

gen.py
```diff
@@ -18,33 +18,35 @@
 from subprocess import PIPE, Popen, call
 try:
     from gi import require_version
-    require_version('Rsvg', '2.0')
+    require_version("Rsvg", "2.0")
```
Foggalong, line 21: Switched `'` for `"` for consistency throughout. The only exceptions are strings for `app_filer` because that uses `"` characters already as part of its content.

gen.py
```diff
@@ -18,33 +18,35 @@
...
     from cairo import ImageSurface, Context, FORMAT_ARGB32
     use_inkscape = False
 except (ImportError, AttributeError, ValueError):
-    ink_flag = call(['which', 'inkscape'], stdout=PIPE, stderr=PIPE)
-    if ink_flag == 0:
-        use_inkscape = True
-    else:
+    # Assigns true if Inkscape found, false otherwise
+    use_inkscape = not call(["which", "inkscape"], stdout=PIPE, stderr=PIPE)
+    if not use_inkscape:
```
Foggalong, line 28: Rather than using distinct `flag` and `use` variables, can just use one with logical `not` converters.

gen.py
```diff
@@ -18,33 +18,35 @@
...
         exit("Can't load cariosvg nor inkscape")
 
 
-parser = ArgumentParser(prog="Numix-core")
-
-# Importing JSON
+# Loading the JSON data file
```
Foggalong, line 32: Load JSON before starting work with the parser

gen.py
```diff
@@ -55,15 +57,17 @@
 
 # User selects the theme
 if not args.theme:
-    exit("Please use --theme argument with "
-         "one of the following: {}".format(", ".join(themes)))
+    msg = "Please use --theme argument with one of the following: "
+    exit(msg + ", ".join(themes))
```
Foggalong, line 62: There were quite a few cases where `format` wasn't needed; standard string concatenation sufficed so uses that.

gen.py
```diff
@@ -97,93 +101,119 @@ def convert_svg2png(infile, outfile, w, h):
...
         png_io.close()
         img.finish()
 
 
-# Only certain icon sizes may be covered
-try:
-    sizes = listdir("icons/{0}".format(theme))
-except FileNotFoundError:
-    exit("The theme {0} does not exists. Please reclone"
-         "the repository and try again.".format(theme))
```
Foggalong, line 112: Because of the prior `try` block in line 42-43 this `try block` (previously lines 107-112) can never actually be trigger. That is, except if the theme directory is deleted between the previous check and this one. Covering that gets us into race conditions though which is a path I'd sooner not go down.

gen.py
```diff
@@ -97,93 +101,119 @@ def convert_svg2png(infile, outfile, w, h):
...
+            # Render icon to resources directory
             drawable_name = icon_name.replace(".", "_").replace("-", "_")
-            output = "{0}/{1}.png".format(theme_dir,
-                                          drawable_name)
-            if path.exists(source):
-                convert_svg2png(source, output, 192, 192)
-            app_filter_content += '\t<item component="ComponentInfo{' + \
-                component_info + '}" drawable="' + drawable_name + '"/>\n'
+            output = "{0}/{1}.png".format(theme_dir, drawable_name)
+            convert_svg2png(source, output, 192, 192)
```
Foggalong, line 139: This was actually a bug! Previously, the icon was added to the app_filter file even if it didn't have an icon created for it.

gen.py
```diff
@@ -97,93 +101,119 @@ def convert_svg2png(infile, outfile, w, h):
...
+
+            # Add icons app_filter data (afd) to the app_filter file
+            afd = '\t<item component="ComponentInfo{{{0}}}" drawable="{1}"/>\n'
+            app_filter_content += afd.format(component_info, drawable_name)
```
Foggalong, line 143: Cleaner representation of what this line actually looks like than splitting it as previous.

gen.py
```diff
@@ -97,93 +101,119 @@ def convert_svg2png(infile, outfile, w, h):
...
+
+# OSX Generation Code
 elif platform == "osx":
     print("\nGenerating OSX theme...")
-    ink_flag = call(['which', 'png2icns'], stdout=PIPE, stderr=PIPE)
-    if ink_flag != 0:
+
+    # OSX generation depends on libicns function png2icns
+    no_icns = call(["which", "png2icns"], stdout=PIPE, stderr=PIPE)
+    if no_icns:
```
Foggalong, line 194: Renamed variable, pretty sure this was just a copy-paste error.

gen.py
```diff
@@ -97,93 +101,119 @@ def convert_svg2png(infile, outfile, w, h):
...
-                convert_svg2png(source, output_png, 1024, 1024)
-                call(["png2icns", output_icn, output_png],
-                     stdout=PIPE, stderr=PIPE)
+            if not path.exists(source):
+                continue
+
+            # Icon output locations
+            svg_icon = "{0}/vectors/{1}.svg".format(osx_dir, output_icon)
+            png_icon = "{0}/pngs/{1}.png".format(osx_dir, output_icon)
+            icn_icon = "{0}/icns/{1}.icn".format(osx_dir, output_icon)
```
Foggalong, line 212: Not gonna lie, I renamed these variables *entirely* so that the final `call` didn't have to be split over two lines.

tests/icon_cache.py
bilelmoussaoui: Are you sure about this? if you load the json file here, it will result in an unordered dict.

Foggalong: The point was more that the `icon_cache` test doesn't need to load the JSON in a special way, the standard load used elsewhere is fine. I didn't test if `object_pairs_hook=OrderedDict` ordered the dictionary, I just assumed it did based on the name.

bilelmoussaoui: sorry I misunderstood what you said then :p it's perfect! 
the object_pairs_hook uses a new object for each child on your json object, as dict don't preserve the order, using OrderDict makes it use the same order as the one on the file

Foggalong: Ahhhh, I get you. That actually makes a lot of sense, thanks! Also if anything else about this pull catches your eye as iffy let me know, I realise it's changing a lot of code you originally wrote

bilelmoussaoui: You can ping me here whenever you need that. Will take a closer look once you think it should be merged ^^

Foggalong:
Something tweaked in the interim has fixed #4779 - my guess is that it was just a case of a character not being being correctly escaped but is as part of some formatting changes.

Foggalong:
It's just suddenly dawned on me that most the dependencies that have been added in `test.yml` and `test-requirements.txt` aren't actually test dependencies, they're _build_ dependencies. We should be able to run the tests just fine without installing anything extra except `flake8` and `jsonschema`.

If we later add a GitHub action to automatically build and push to `numix-icon-theme-*` then those will need them other dependencies, but that will also have to be a separate workflow since we wouldn't want it to trigger `on: pull_request`

PR #4783, opened 2019-02-15
Discreete Linux Icons (2)

Foggalong:
Follows on from #4737 because apparently I can't resolve merge conflicts locally to save my life.


Foggalong:
@palob This is now (finally) ready to be reviewed design/icon wise :smile:

Foggalong:
@palob This is now (finally) ready to be reviewed design/icon wise smile :smile: EDIT: meant to add, here are previews of the redesign of the GnuPG since it changed existing designs.

Old:
2019-02-15-181303_477x103_scrot
New:
2019-02-15-181328_475x111_scrot

palob:
The GnuPG icons look cool from distance already.

palob (changes requested):
+ In ``cryptobox-search`` the looking glass is slightly misaligned. For the sake of simplicity it should be implemented as to circles on top of each other (it's a ring with circle underneath currently).
cryptobox-misalign

+ ``cryptobox-wizard`` should have all shadow paths merged.

+ The baseplate gradient of the gnupg icons ever so slightly changes the saturation.

palob:
Looks like the db entry for ``mat`` is still duplicated yet it passes Travis?

Foggalong:
@palob I believe duplicated entries isn't something we currently properly test for. Added an issue to create a test for that as part of #4780.

Foggalong:
@palob I've made all the changes, plus some some minor alterations to the Cryptobox Search design, apart from the GnuPG gradient change. I don't know whether it's a bug with my Inkscape version or a property of those particular HSL values but I can't change the top saturation to 167 without Inkscape automatically correcting it back to 161, even when I start from scratch with a new baseplate and gradient. Do you experience this issue?

palob:
That's probably just due to the non-bijecfiveness of the HSL-RGB conversion. With H and L unchanged both saturation values (and the ones in between) will result in the same RGB colour value.

Foggalong:
Ahh fair fair, I've not seen that crop up over such a big gap before. Might be something to due with the purple hue. But anyways ready for review if that's the case :)

palob (approved):

Foggalong:
Thanks @palob! Will see about this getting merged into upstream so that Discreete can just use the standard Debian package going forward rather than their own fork

palob:
That'd be great. There are still some distros which use forks of our theme instead of contributing here (e.g. Manjaro, I believe). As a result they miss out on any of our further developments.

PR #4790, opened 2019-02-24
Icons for Toggl

palob:
https://github.com/toggl/toggldesktop

toggldesktop
toggldesktop-ci
toggldesktop-sq


Foggalong:
As a daily Toggl user myself I've been wondering when this would come up. I've had the thought that maybe we should intentionally deviate from the upstream design for this icon because it so closely matches or "shutdown" design. I'm not sure what that deviation would look like, possibly something closer to a stopwatch, but I'd be interested to see whether a move in that direction would be possible.

Foggalong (changes requested):
As a daily Toggl user myself I've been wondering when this would come up. I've had the thought that maybe we should intentionally deviate from the upstream design for this icon because it so closely matches or "shutdown" design. I'm not sure what that deviation would look like, possibly something closer to a stopwatch, but I'd be interested to see whether a move in that direction would be possible. Happy to be convinced otherwise, but that's why it's marked as "for changes" for now

palob:
Well, tbh I don't exactly know what Toggl does. The 0/1 button probably isn't all that good of a metaphor.

Foggalong:
It's an app for time tracking, designed I think primarily for self-employed peeps and people who use self-employed peeps to make invoicing easier and more accurate. I started using it in 2nd year of uni following a recommendation from [Cortex](https://www.relay.fm/cortex) to work out how well I was managing to keep a university-work-to-extra-curricular balance.

palob:
Ah, being a prolific procrastinator myself, that's probably why I dont know it :-)
So, something stopwatch-/hourglass-/clock-like and a red background?

Foggalong:
@palob Yeah, that would work!

palob:
Some drafts, the last one might be a bit too fine-grained.

toggldesktop-0
toggldesktop-00
toggldesktop-1
toggldesktop-2
toggldesktop-3

Foggalong:
Go with design 4 - Toggl is a timer in the sense it counts up time (like a stopwatch) rather than down time (like an egg timer).

palob:
You mean the digital one?

Foggalong:
Aye! It's a good design :)

bilelmoussaoui:
@palob you can try to reach upstream devs if they want a logo based on the icons you have made. They look great!

palob:
Pushed
toggldesktop-ci
toggldesktop-sq

Foggalong (approved):

PR #4791, opened 2019-02-24
Icons for Tauon

palob:
https://github.com/Taiko2k/TauonMusicBox

tauonmb
tauonmb-ci
tauonmb-sq


Foggalong (changes requested):
This design doesn't translate well to being used as an entire baseplate because it's emulating a disk record which looks weird on square. Would be fine if it was maybe a purple disk on a black background for both? Or a black disk on a purple background

palob:
How about only changing it for Square?
tauonmb-sq

Foggalong:
@palob Looks great! I'd be leaning toward changing it for both; I've been toying with the idea of starting to move away from full-baseplate designs in Circle to have greater consistency between it and Square. Firefox is the one that got me thinking on this because the FF Focus icon looks much nicer in terms of being a centred symbol than stretched to fill the space.

palob:
Pushed
tauonmb-ci
tauonmb-sq

Foggalong (approved):

PR #4798, opened 2019-02-26
technical optimisations

palob:
optimisations in order to reduce filesize and rendering overhead.


Foggalong (changes requested):
In Almanah the needs to be some tweaking to the lock and latch so that they have a slight overlap. This prevents glimpses of red showing through. There are other icons in this PR where this occurs, but Almanah is the only one where this PR introduced it.

Small thing, but Picard should also have an extra shadow for the music note (but not an extra the top tag, since I'm assuming they're made of thin card).

palob:
Was working on a more thorough picard redesign when bloody Inkscape crashed me. Overall ``0.92.x`` has been a very crashy series for me. Will start running the master version. Can't be worse and I'll be rid of GTK 2 there.

Foggalong:
@palob No worries, there's no rush on it!

palob:
They've got a new icon anyway https://picard.musicbrainz.org/.

palob:
Here's the new picard icons I'm gonna push.
picard-ci
picard-sq

Foggalong:
Those are brilliant!

palob:
Pushed together with these two:
almanah-ci
almanah-sq

Foggalong:
Something about that lock design doesn't look right, I can't work out if it's the lack of 3D perspective (this icon is naughty in that it uses one, which make make elements which overhang look off) or the fact that the buckle isn't vertically centered.

palob:
3 more drafts:
almanah-2
almanah-3
almanah-4

Foggalong:
Big yes on 3, much better design :heart_eyes:

palob:
Pushed :+1:

Foggalong (approved):

PR #4799, opened 2019-02-26
pycad rework. Fixes #4765

### LINKED ISSUE #4765: PyCAD Redesign
The current symbol with thick black outlines doesn't fit in especially well with the theme. A complete redesign is one possible fix, or else just reducing the prominence of the outlines. This is also an application which may have had a redesign since the icon was originally created.

palob:
pycad-ci
pycad-sq


Foggalong:
Is there any way to reform these shapes so it's not a 2px gap in the middle?

palob:
Only at the cost of either alignment or symmetry.
But maybe a 25/27px high symbol could solve the issue.

Foggalong:
@palob That's a possibility, was just curious if it could be tweaked because as is the gap catches the eye

palob:
pycad-ci
pycad-sq

palob:
The non-centered-ness isn't noticeable for me. This means the same should be done for all Python icons where currently the gap is located at  a ``x.5`` Y value.

Foggalong:
Perfect, not noticeable for me ether. I'll merge this in but yeah we should probably tackle that also. The Python icons generally are not as nice looking as I remember them being.

Foggalong (approved):

PR #4812, opened 2019-03-10
Set sublime-text-3 icon as default

jaimelive:
Since Sublime Text 3 has been released in 13 September 2017, and most of the user base is on that version, I propose setting sublime-text-3 as the default icon.
Desktop entries use `Icon=sublime-text`, so sublime-text.svg needs to be moved to sublime-text-2.svg and sublime-text-3.svg symlinked as the previous.


Foggalong (approved):
Good call, had a check and the packaging for 3 all seems to be using `sublime-text` now so all good. Added an extra commit just so this doesn't needlessly break uTouch and Shine.

PR #4813, opened 2019-03-11
Symlink for Cast to TV

Rafostar:
Can symlink for [gnome-shell-extension-cast-to-tv](https://github.com/Rafostar/gnome-shell-extension-cast-to-tv) be added?

### DIFF
--- a/data.json
+++ b/data.json
@@ -11553,6 +11553,7 @@
         "linux": {
             "root": "mkchromecast",
             "symlinks": [
+                "cast-to-tv",
                 "gnomecast"
             ]
         }


palob:
Since this is a Shell extension, where is this icon shown?
Please add the name of the original icon and how it looks like.

Rafostar:
After selecting type of media you want to cast, file chooser dialog gets opened to make selection. This dialog is its own Gtk app that appears in app dock like any other opened app. Currently no icon is set in app code, so it looks like this:

no-numix-icon

And when I specify name of "mkchromecast" or "gnomecast" icon (numix provides those icons). The opened app can look like this:

numix-icon

But I would not like to specify other app icon name in code if creating this symlink would be possible. Then I could set the app to use its own icon from icon theme and do a fallback to generic app icon if theme does not provide it.

Rafostar:
I have written icon name that I would like to use inside `data.json` in this PR. The icon should look exactly like the one used for mkchromecast so symlink is enough.

Foggalong (approved):

Rafostar:
Thank you very much!

PR #4818, opened 2019-03-15
Icons for Dust3D

palob:
https://github.com/huxingyi/dust3d

dust3d
dust3d-ci
dust3d-sq


Foggalong (changes requested):
As much as a camel is a camel, I don't think we should use the symbol we use for Perl for this.

palob:
Pushed
dust3d-ci
dust3d-sq

Foggalong:
This gave me a giggle when I opened it, love the low poly camel :+1:

Foggalong (approved):

PR #4821, opened 2019-03-15
Extra VS Code Symlinks

Foggalong:
Symlinks for #2964 and microsoft/vscode#62650.


Foggalong:
I'll try not to think too much about why that's the ordering that works.

PR #4857, opened 2019-04-28
Update icons for Haguichi to match the current logo (resolves #3709)

### LINKED ISSUE #3709: Haguichi icon improvement
It will be nice to match the icon with the new logo.

image

Description: `Haguichi is free and open source software licensed under GPLv3+. Written in Vala and using GTK+ as widget toolkit, Haguichi provides a user friendly GUI to ...`

Website: `https://www.haguichi.net/news/release130/`

ztefn:
| --- | :-- | :-- |
|   |   |  |


palob (changes requested):
Thanks for contributing. There's a few small things left.

+ I'd say use the slighltly softer red from the website for the baseplate gradient, meaning L162 at the bottom
+ The white path should be coloured ``f9f9f9ff`` instead of plain white. It is the brightest we use.
+ The white path has got an unneeded note.
+ The green path maybe should be toned down a bit to, say S240 (at the moment it is fully saturated).
+ All nodes should have integer coordinates or at least ``.5``ones. Should be doable without distorting the icon significantly.

ztefn:
@palob I followed your suggestions except for the _slighltly softer red_ because that would turn it visually into pink instead (which is not what I'm intending).

palob:
Maybe we misunderstood? Just ever so slightly lighter like the top bar here https://www.haguichi.net/

The upper node of the green path shouldn't exceed Y23.

ztefn:
> Maybe you misunderstood. Just ever so slightly lighter like the top bar here https://www.haguichi.net/

Yes, but if I would use that as bottom color (instead of top), the top color needs additional L10. Thus turning it into pink:
Screenshot from 2019-04-29 23-58-09
Screenshot from 2019-04-29 23-57-43


> The upper node of the green path shouldn't exceed Y23.

The lower path has little green. That node at X21.618 Y23.177 is exactly where it should be. I'm not skewing my logo any further.

Foggalong:
Happy to compromise on the background and that node placement; the middle one which is off grid has less of an impact alignment wise than the others so it'll make little difference when it's seen in practice. Thanks again for contributing this, it's not often we have the upstream dev come in with an icon!

Foggalong (approved):

PR #4869, opened 2019-05-12
Flatpak symlinks

palob:


data.json
```diff
@@ -6099,7 +6102,7 @@
                 "adobe-flashplayer",
                 "AdobeFlash",
                 "adobeflashplugin",
-                "com.Adobe.Flash-Player-Projector",
+                "com.adobe.Flash-Player-Projector",
```
Foggalong, line 6105: Is there a reason this is a rename rather than a new icon?

palob: It's not an addition, the old one was a typo.

Foggalong (approved):

PR #4872, opened 2019-05-24
Icons for Pace

palob:
https://gitlab.com/softwareperonista/pace

pace
pace-ci
pace-sq


Foggalong (changes requested):
I like the idea of this as a stylistic Pacman icon, but not sold on it for this app because it's seems like a redesign. Did you have any joy with the foot design?

palob:
I did opt against the foot because it might suggest a relation to GNOME which does not exist.

Foggalong:
That's actually a very good point, especially with the blue colour. Will merge this and then if upstream take issue with can discuss this in more detail with them then

Foggalong (approved):

PR #4892, opened 2019-06-09
Refreshed Godot design

### LINKED ISSUE #4584: Update Godot Engine icon
| --- | :-- | :-- | :-- |
| `Godot Engine`| `godot` | `godot` | original |

Description: `Multi-platform 2D and 3D game engine.`

Website: `https://godotengine.org`

The current icon is outdated, using the old Godot Engine logo.

Foggalong:
Upstream branding update, but also switched from cutouts to symbols on symbols. Fixes #4584


palob (commented):
Works well but the symbol is a bit small imo. Did you try a bigger size?

Foggalong:
I tried but it's difficult to keep the teeth proportions looking okay at 24-26 px. I can have another bash and report back with some examples

Foggalong:
I've had a play with this but haven't yet found a 24px+ version which doesn't end up with an oddly warped mouth. Will keep trying but won't be offended if anyone else swoops in to have a bash.

palob:
Is it the decision between a 1px or a 2px wide mouth stroke where something in between would be ideal?

palob:
Currently in this PR:
godot0

I had a bash at it, 2 variants:
godot1
godot2

palob:
The last one is the closest to the original icon but I like it the least. Maybe it could be bigger still?

Foggalong:
The chin of the middle looks a bit fat but otherwise okay. The last is better, but those eyes make it look very off. It's an odd one to get working right with the proportions of the face

palob:
This is variant 3 with asquint eyes similar to the original icon.
godot3

palob:
Here's a bigger variant with the original icon on the left. Mouthline shouldn't be that close to the original probably for the sake of crispness.

godot4

Foggalong:
[This one](https://github.com/numixproject/numix-core/pull/4892#issuecomment-505599353) is starting to look too large I think, though the mouth did come out really well. The eyes on that and [this one](https://github.com/numixproject/numix-core/pull/4892#issuecomment-504065963) also look better but make it look a little cross eyed because the pupil is so much smaller than the whites. Maybe [this](https://github.com/numixproject/numix-core/pull/4892#issuecomment-504065963) symbol size and shape but with the eyes from [variant 1 here](https://github.com/numixproject/numix-core/pull/4892#issuecomment-503983553)?

palob:
Looks like this
godot-final

Foggalong:
@palob That's the one, great work! What an unexpectedly difficult icon this turned out to be

palob:
Pushed!

PR #4905, opened 2019-06-15
Icons for Vorta

palob:
https://github.com/borgbase/vorta

com borgbase Vorta
com borgbase Vorta-ci
com borgbase Vorta-sq


Foggalong (changes requested):
Not a massive fan of the green for this. Is there a yellow or red which would work well?

palob:
I took the green from the Borg typeface but that's a bit far-fetched anyway so I switched to this:
com borgbase Vorta-ci
com borgbase Vorta-sq

Foggalong (approved):

PR #4927, opened 2019-07-04
Icons for Zrythm

palob:
https://www.zrythm.org/en/

zrythm
zrythm-ci
zrythm-sq


Foggalong (changes requested):
Can we make the orange a little more saturated and the baseplate more red-brown? The colour feels very washed out as is

palob:
Pushed these
zrythm-ci
zrythm-sq

Foggalong (approved):

PR #4935, opened 2019-07-11
Icons for usbguard

palob:
https://github.com/USBGuard/usbguard

usbguard-icon
usbguard-icon-ci
usbguard-icon-sq


Foggalong (changes requested):
Can we get a dark navy blue for the symbol colour instead?

palob:
This?
usbguard2

Foggalong:
Yeah that's good with me! Just noticed, a wee bit naughty that they're using the Google G but not our place to change that I don't think :P

palob:
Pushed for both.

Foggalong (approved):

PR #4941, opened 2019-07-16
Icons for Raven Reader

palob:
https://github.com/hello-efficiency-inc/raven-reader

raven-reader
raven-reader-ci
raven-reader-sq


Foggalong:
This symbol reminds me a lot of the rabbit-duck illusion, but I don't think it quite works here. Is there a way to incorporate in the swoosh at the bottom like in the original?
 
250px-Kaninchen_und_Ente svg

Foggalong (changes requested):
This symbol reminds me a lot of the rabbit-duck illusion, but I don't think it quite works here. Is there a way to incorporate in the swoosh at the bottom like in the original?
 
250px-Kaninchen_und_Ente svg

palob:
Or  maybe simply the RSS icon recoloured?

Foggalong:
@palob Yeah I'd be good with a recolour in this instance :)

palob:
Pushed these
rss-c
rss-ci

Foggalong (approved):

PR #4942, opened 2019-07-16
Icons for Ultragrid

palob:
https://github.com/CESNET/UltraGrid

ultragrid
ultragrid-ci
ultragrid-sq


Foggalong (changes requested):
The Circle icon just about gets away with having the symbol off center, the Square one does not. I'm not sure this could be rectified without changing the shapes though :confused:

palob:
As a matter of fact the symbol on the Square icon was misplaced 1px to the right.

Looks like this now
ultragrid-sq

Foggalong (approved):

PR #4956, opened 2019-08-03
Icons for OpenComic

palob:
https://github.com/ollm/OpenComic

opencomic
opencomic-ci
opencomic-sq


Foggalong (approved):
Liked this one, but just switched out the `#f9f9f9` for a more yellow `#fff0d4`. Draws from the off white used in the original and it's also slightly nicer on the eyes

PR #4974, opened 2019-09-25
Icons for Obfuscate

palob:
https://gitlab.gnome.org/World/obfuscate

Based on an older iteration of the original icon

obfuscate
obfuscate-ci
obfuscate-sq


Foggalong (changes requested):
This is actually a nicely designed icon, but I'd like us to get something in which is based on the current icon given they're so different. Want to minimise any possible confusion there

palob:
The current official design is hard to get recognisable at smaller sizes.

How about a face with a censor bar across? Might suggest this was for photo editing though.

So basically something which shows letters/numbers on one side turning into blocks on the other side.

Foggalong:
@palob Yeah, doesn't necessarily have to be exactly the same design. Could just be 4 numbers with two as is and 2 blocked?

@bilelmoussaoui Do you have any input on this? Realised looking at the Gitlab that it's actually your app!

bilelmoussaoui:
This was the very first iteration of the icon and it doesn't really represent what I want to achieve with the app, especially in the future. Probably either using the 2nd iteration if you want something that could be recognizable at smaller sizes.

@Foggalong Yes it's my app :P

Foggalong:
Is the second iteration different from the current icon? I've not come across that when looking online

bilelmoussaoui:
@Foggalong https://gitlab.gnome.org/World/obfuscate/blob/cade72b5d62c7e2275ea0d63ba0e34b5035b4450/data/icons/com.belmoussaoui.Obfuscate.svg

Foggalong:
@bilelmoussaoui Oooh brilliant, that design will lend itself much better to scaling! Thanks :grin:

palob:
Pushed this direct representation of the new upstream design
obfuscate
obfuscate-ci
obfuscate-sq

palob:
alternative idea

obfuscate-alt

Foggalong:
I love the pushed design _but_ also that censorship icon is a really cool idea. Keep that around somewhere (even if it's just pushing it to Numix Labs) because it'd be a shame to let it go to waste

Foggalong (approved):

PR #4987, opened 2019-10-07
Add missing Celluloid symlink

palob:


Foggalong:
why would they do this :cry:

bilelmoussaoui:
The first icon name is wrong, it's not a valid app-id and was probably changed before the release is out and can safely be removed

Foggalong:
Curious, why is `io.github.Celluloid` not valid?

palob:
It might as well have been a typo introduced by me but there are probably a lot of invalid app ids out there with Flatpaks distributed outside of Flathub.

Foggalong:
Eh, we'll leave it in then just in case. No sense in spending time hunting down whether every icon line in the JSON is something that could still occur in the wild.

PR #4989, opened 2019-10-27
Flatpak symlinks

palob:


palob:
Oh, wasn't aware there's already one open, sorry.

Foggalong (approved):

PR #5001, opened 2019-11-09
Circle: shift shopping bags up by 1px

palob:
+ gradient/alignment fixes


Foggalong (approved):

Foggalong:
Good catch on the Debian circle shadow being misaligned!

PR #5018, opened 2019-12-27
Icons for Setzer

palob:
https://github.com/cvfosammmm/Setzer

upstream icon redesign discussion
https://gitlab.gnome.org/cvfosammmm/setzer/issues/2

setzer
setzer-ci
setzer-sq

Alternatives
setzer0
setzer0c
setzer1
setzer4
setzer5
setzer7


Foggalong (changes requested):
I'm not sold on any of these unfortunately, they're far enough away from both the original icon and the use of the app that I'm not sure it'd be recognised.

palob:
*Setzer* however I think is supposed to mean typesetter.
Seeing the upstream discussion the bird might only be a temporary feature of the icon.

palob:
Maybe a typewriter?

palob:
Pushed this. Better or too detailed?

setzer2-sq

setzer2-ci

Maybe without the lines on the sheet?

setzer222

palob:
Another variant
setzer3-ci
setzer3-sq

Foggalong:
The typewriter is brilliant! Pushed that one :heart:

Foggalong (approved):

PR #5033, opened 2020-01-16
Icons for wxlame

palob:
https://github.com/cfgnunes/wxlame

wxlame
wxlame-ci
wxlame-sq

Based on the lame logo which this is a frontend to.
Maybe a tad too small/fine-grained.
Could try something like
la
me


palob:
Hmm, not that recognisable either (+ tad too big)
wxlame-2

Foggalong (changes requested):
I think with this one we'd be better using the speaker design with a green baseplate rather than the base-tool logo

PR #5050, opened 2020-02-13
Icons for guiscrcpy

palob:
https://github.com/srevinsaju/guiscrcpy

guiscrcpy_logo
guiscrcpy_logo-ci
guiscrcpy_logo-sq


Foggalong (changes requested):
@palob I can't request your review because you're OP, but what do you think to adding some background symbols like this? The symbol was looking unusually long and thin compared to the icons around it when I was reviewing. More than happy for you to tinker with the design and colouring from there but I do think it needs something more to widen it out a little.

guiscrypy-square guiscrypy

palob:
That's lovely, please merge it!

Foggalong (approved):

PR #5060, opened 2020-02-27
Ubuntu Kylin logo

palob:
ubuntu-kylin-ci
ubuntu-kylin-sq

Pushed a new Ubuntu logo as well but I'm not sure if it's better.
*old*
ubuntu-before-ci
ubuntu-before-sq

*new*
ubuntu-remade-ci
ubuntu-remade-sq


Foggalong (changes requested):
I'm struggling to find the conversation now on GitHub, but it was a concerted decision to replace a design similar to the Ubuntu one you've suggested with what we have now. The reasoning if I'm recalling correctly was the awkwardness of aligning the design without the circle and it looks a little weird generally on Square.

I'm less sure of that decision now looking at your designs, but the symbols do need shifting a pixel or two to the right in Square to counter the triangle-like off-centredness illusion.

palob:
Looks like this (pushed)
ubuntu-kylin-sq
ubuntu-remade-sq

Foggalong (approved):

PR #5069, opened 2020-03-08
Icons for Gamma

palob:
https://gitlab.com/hamadmarri/gamma-text-editor

gamma
gamma-ci
gamma-sq


Foggalong (changes requested):
I think you could get away with making these lines at least twice as thick as they are now

palob:
Pushed
gamma-ci
gamma-sq

Foggalong (approved):

PR #5075, opened 2020-03-13
redesign for preferences-desktop-user

palob:
old

user-before

new
user-after-ci
user-after-sq


Foggalong (changes requested):
I'm gonna take a radical stance and wontfix this issue; I'm not super keen on the changes, though I do see where you were coming from in updating it; the original is a bit messy both technically and in design. If we' are changing this icon though I think we should do a slightly wider ranging change and add skin/gender options a la emojis for this and other generic application icons which use this as a template.

That quite a significant job though, so I'll make another issue for it. Doesn't feel right to have that as a continuation of this.

PR #5076, opened 2020-03-14
dedicated icons for JOSM

palob:
josm
josm-ci

josm-sq

Too many details on the map?


Foggalong:
Not at all, I think this works very well!

Foggalong (approved):

PR #5089, opened 2020-03-17
templates: remove confusing opacities

palob:
so you can't forget to remove it later. Colours adjusted accordingly.
+optimisation


Foggalong (approved):

Foggalong:
I have no idea why at the time I thought that was a good idea :sweat_smile:

PR #5091, opened 2020-03-17
optimisations

palob:
+symbol-on-symbol shadows for ``system-software-...``


Foggalong (changes requested):
I'm not convinced by the symbol on symbol shadows for snapshot; in circle in particular it makes the icon look quite cluttered and I don't think adds anything (unless the eyes are actually standing out from the face?). Generally good though, I like where soss have been added elsewhere!

palob:
Sorry, needed to revert, 2 files were edit which need into the other style cleaning PR.

palob:
Should be safe now.

Foggalong (approved):

PR #5100, opened 2020-03-20
Icons for Electronic Super Joy

### LINKED ISSUE #3232: Icon for Electronic Super Joy (Steam)
**Game:** Electronic Super Joy
**AppID:** 244870
**Icon:** 
Electronic Super Joy
**Banner:**
Electronic Super Joy
**Notes:** I believe the banner could be used more for reference than the icon, it seems as though the character to the left of the banner is the same as the character in the pixel-art icon. Maybe following the same style as the icon while using the high quality of the character on the banner would be the way to go.

maxtron95:
Fixes #3232

electronic-super-joy-ci svg
electronic-super-joy-sq svg


palob (commented):
In oder to get rit of the fine outline where those "exhaust path and the baseplates share the same edge, can you please do the following?
+ duplicate the "exhaust" path, remove all nodes but the outer to at the baseplate edge.
+ straighten the inwards handles of the two remaining nodes and extend the path outwards so it looks something like this

+ subtract this path from the baseplate

maxtron95:
@palob What do you mean by subtracting the path from the baseplate? Do you mean duplicating the baseplate on top of the path and then intersecting them?

palob:
You need to select both the new path and the baseplate, have the new yellow path above the baseplate and perform *Path -> Difference*. No need to duplicate the baseplate.

By the way this is what I was referring to with "exhaust" path.


The baseplate will look like this in the end (made the "exhaust" transparent for demonstration).



The Square baseplate will have cut a bigger area.

maxtron95:
@palob I see what you are talking about. Thanks.

palob (approved):
That's it, thank you.

palob:
I don't quite get why the Travis CI build is failing here. db order is correct, isn't it?

maxtron95:
Yeah, It placed the electronic-super-joy icon after electron-cash in my file manager.

palob:
Ah, the new DB entry clashes with ``electron-mail`` now in master.

maxtron95:
@palob So what do you think I should do about it?

palob:
Pull the chanhes from master but don't know if that's needed.
We could just fix the order in a commit after merging.

PR #4867, opened 2019-05-12
baseplate gradient fixes

palob:
Fixed a few baseplates where the gradient oviously exceeded the 10L span.

PR #5103, opened 2020-03-20
Icons for Doki Doki Literature Club

maxtron95:
https://ddlc.moe/

43a7dff3d95905026e6d00f45458a2d7_icon

doki-doki-literature-club-ci svg
doki-doki-literature-club-sq svg


palob (changes requested):
+ Could we get a starker baseplate-symbol contrast by means of a brighter background colour closer to the one in the original icon?
+ The dotted line has many nodes which can be removed without altering the shape too much.
There's a small glitch at the bottom, should be like this (or better with the node right at the cross)



+ The outer rim probably needs to go, we've removed those from our icons, e.g. [here](https://github.com/numixproject/numix-core/pull/4867/files?short_path=a1b103f#diff-a1b103f3cede66f0710e0b3c58f2ff54)
On the other hands we still have rims on the borderlands icons.

maxtron95:
@palob I pushed these.
ddlc-ci
ddlc-sq

maxtron95:
@palob Do you think these icons are now approved since I changed them?

palob (changes requested):
Can you please move the symbol down by 1px for circle and turn 
```
style="color-rendering:auto;color:#000000;dominant-baseline:auto;fill:#1284e8;font-feature-settings:normal;font-variant-alternates:normal;font-variant-caps:normal;font-variant-ligatures:normal;font-variant-numeric:normal;font-variant-position:normal;image-rendering:auto;isolation:auto;mix-blend-mode:normal;shape-padding:0;shape-rendering:auto;solid-color:#000000;text-decoration-color:#000000;text-decoration-line:none;text-decoration-style:solid;text-indent:0;text-orientation:mixed;text-transform:none;white-space:normal"
```
into
```
style="fill:#1284e8"
```

palob (approved):

Foggalong (approved):
I liked this one, just after the border was removed the symbol was a little too small. Upped the size a bit and made the baseplate 5L darker so that it's not quite so white at the top. Definitely the right call to make it less saturated than the original submission though.

PR #5108, opened 2020-03-21
Icons for Droid Assault

maxtron95:
https://store.steampowered.com/app/219200/

droid-assoult

droid-assault-ci svg
droid-assault-sq svg

### DIFF
--- a/data.json
+++ b/data.json
@@ -5285,6 +5285,11 @@
         "linux": {
             "root": "drill-search-gtk"
         }
+	},
+	"droid-assault": {
+		"linux": {
+			"root": "steam_icon_219200"
+		}
     },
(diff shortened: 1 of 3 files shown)


palob (commented):
Icons are ready for merge designwise. Symbol is just a bit big for a non-nearly-circular design. Either we make an exception (which I think we did for some complicated game icons) or the left feet could be moved up 1px and the other one could be brought in line with it to get a height of 26px.

Foggalong (approved):
Agree re it being too big, just submitted a quick PR to move the right two feet up by 1px each, shorten all three by 1px, and then recentre the whole symbol again. Cracking design work though @creepertron95!

Issue #4323, opened 2018-04-20
Icon Request : MultiMC


founderio:
| Application name | Icon name | Desktop file name | Original icon | Theme |
| --- | :-- | :-- | :-- | :-- |
| MultiMC | `multimc` | `multimc.desktop` | ![MultiMC](https://i.imgur.com/Rha4tMK.png)| Numix-Square |

This issue was automatically created using [IconRequests](https://github.com/bil-elmoussaoui/Icon-Requests).

maxtron95:
@palob Should this issue be closed?

founderio:
I've updated and the icon is there now, so I'd say yes :)

PR #5111, opened 2020-03-22
Icons for MultiMC

Tank-Missile:
| --- | :-- | :-- |
| multimc-original | multimc-circle | multimc-square |


Tank-Missile:
One of the MultiMC developers suggested I add grass to the top of the icons to make them look more like the original, but it may make them look cluttered. I'm not too sure.

palob (changes requested):
Hi, thanks for your contribution. You're right, I think the grass would clash with how we usually design our icons.

There are a few issues which need to be adressed.
+ We don't use solid colours for our baseplates but rather vertical gradients. Starting from our ``48.svg`` files in the template folders you'll find these two nodes left to the icon (blue line in between).
nodes
This is where you can set the 2 stop colours of the gradient. In the HSL colour space the upper noder should be 10L (for Inkscape <1.0) lighter compared to the one at the bottom.

+ The symbol is a bit too big. Width should be no more than 26px or 28px.

+ The symbol needs to be aligned with the pixel grid (to be toggled with the ``#`` key, otherwise the icon will appear blurry. The ∞ symbol needs to have integer px width and height dimensions and the stroke width must be an integer number of pixels.

+ The symbol needs to have a drop shadow. This is achieved by duplicating the symbol, turning it black with 10% opacity, moving it below the symbol and shifting it 1px to the right and 1px down (only 1px down for Square)

+ Please save your icons as *Optimised SVG* with options as shown here https://github.com/numixproject/numix-core/wiki/Optimise-Options

This sounds like a lot, don't be discouraged.

palob:
Fixes #4323.

Tank-Missile:
I think I've made the changes you requested. I'm very new to Inkscape. In fact, this is my first project in it. Hopefully I didn't miss anything.

palob (changes requested):
Thank you for further improving this.
+ Exceptions aside symbols+shadow a sit 1px higher in Square icons compared to the position in Circle icons.
+ The total size of the symbol is correct now
+ The stroke width however needs to be 2px or 3px. The symbol needs to be redone from scratch.
Choose the Bezier tool in the toolbox (icon shows a pencil+straight line) and put nodes at the following positions
nodes
Change the stroke width to 2px. Turn the colour into something transparent so you can use the old symbol as template. Turn the corner nodes into smooth nodes.
nodes2

Adjust the node handles to omething like this
nodes3
and convert the stroke to a path (*Path -> Stroke to Path*). Set the correct colour, remove the old symbol, and duplicate for the shadow.

Tank-Missile:
I redid the symbol as you asked. I appreciate the directions for a new user such as I.

palob (commented):
That's it, thank you.

Unfortunately though, there is some nasty noise in the icon which only can be removed manually.

Can you please open the 2 icons in a text editor and replace
```
style="color-rendering:auto;color:#000000;dominant-baseline:auto;font-feature-settings:normal;font-variant-alternates:normal;font-variant-caps:normal;font-variant-ligatures:normal;font-variant-numeric:normal;font-variant-position:normal;image-rendering:auto;isolation:auto;mix-blend-mode:normal;opacity:.1;shape-padding:0;shape-rendering:auto;solid-color:#000000;text-decoration-color:#000000;text-decoration-line:none;text-decoration-style:solid;text-indent:0;text-orientation:mixed;text-transform:none;white-space:normal
```
by
```
style="color:#000000;opacity:.1"
```

and 
```
style="color-rendering:auto;color:#000000;dominant-baseline:auto;fill:#75b54b;font-feature-settings:normal;font-variant-alternates:normal;font-variant-caps:normal;font-variant-ligatures:normal;font-variant-numeric:normal;font-variant-position:normal;image-rendering:auto;isolation:auto;mix-blend-mode:normal;shape-padding:0;shape-rendering:auto;solid-color:#000000;text-decoration-color:#000000;text-decoration-line:none;text-decoration-style:solid;text-indent:0;text-orientation:mixed;text-transform:none;white-space:normal"
```
by
```
style="fill:#75b54b"
```
?

You can simply do this via the *Search and Replace* functionality of your text editor, invoked by ``Ctrl+H``.

palob:
@Foggalong @creepertron95 Is suspect this infamous SVG noise (see comment above) originates from stroke to path conversions. Will make it easier to hunt for. Will check existing icons and my PRs later.

Tank-Missile:
I've removed the noise from the icons as you asked. What exactly is this noise anyway? I'm no expert on the svg format. All I know is that they display in vectors rather than pixels.

palob (approved):
That's it. Thank you for pulling it through.

palob:
This noise is properties defaulting to standard values. It's all about stylings you can apply but we don't need. You can do pretty sophisticated stuff with SVG.

PR #5114, opened 2020-03-22
Icons for mapton

palob:
https://github.com/trixon/mapton
mapton
mapton-ci
mapton-sq

Wondering if this is viable or I should stick to a standard earth design.


maxtron95:
Well the description of the program is literally "Some kind of map application" on the github repo.

maxtron95:
@palob You might have to change that design, due to #5153.

palob:
But I for one haven't come up with an idea how to fix that one yet.

maxtron95:
@palob Maybe a map design might work here?

Foggalong (changes requested):
I like this _idea_ of a design, but I'm blocking icons like this until #5153 just is addressed to just decrease work down the line. I think this icon though highlights that maybe some of these redesigns would be beneficial to keep a design rooted in real-world continents rather than fantasy continents.

palob:
Alternatively mapton could be implemented as a symlink in #5350.

palob:
Since this is "some kind of map application"...

mapton2

I'm unsure about the buttons, what do you think @Foggalong ?

PR #5115, opened 2020-03-22
Icons for feathernotes

palob:
https://github.com/tsujan/FeatherNotes

feathernotes
feathernotes-ci
feathernotes-sq


maxtron95:
@palob Perhaps we can also include an icon for Featherpad as well. (https://github.com/tsujan/FeatherPad)

palob:
Right now it is implemented as symlink, so maybe just this icon recoloured, or the feather on top of a sheet. It is not easy to distinguish between featherpad and feathernotes.

Foggalong (approved):

palob:
@creepertron95 I'll try something for featherpad next week unless you plan to do so.

palob:
Two drafts for featherpad which I don't exuberantly like:

featherpad
feathernotes1
feathernotes2

Maybe it needs the original feather which then should be pushed to feathernotes as well.

PR #5116, opened 2020-03-22
Icons for Titan Attacks

maxtron95:
https://store.steampowered.com/app/203210/

titan

titan-attacks-ci svg
titan-attacks-sq svg

### DIFF
--- a/data.json
+++ b/data.json
@@ -18670,7 +18670,7 @@
             "root": "tintii"
         }
 	},
-	"titan-attacks" {
+	"titan-attacks": {
 		"linux": {
 			"root": "steam_icon_203210"
 		}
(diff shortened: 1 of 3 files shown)


palob (changes requested):
After ``"titan-attacks"`` in data.json a ``:`` is missing.
Usually we try to avoid gradients on new icons but I think it's okay here.

palob (approved):

Foggalong (changes requested):
Can we switch up the eyes? Gradients in symbols generally aren't within guidelines and I'm not sure it's adding much to the design here. If you _do_ want some sort of colour change I'd be curious how a two-tone colour change at a specific pixel would look.

palob:
Aside from that I think the eyes could be brighter.

palob (approved):

Foggalong (approved):

PR #5117, opened 2020-03-23
Icons for F1 2015

maxtron95:
https://store.steampowered.com/app/286570

f12015

f1-2015-ci svg
f1-2015-sq svg


palob (commented):
You don't neccessarily need to put the jag nodes at integer X coordinates as there are diagonal lines involved anyway. I think it will look better if you keep a straight line for the  jag endings as in the original icon. But it looks better how your jags begin further to the right.
jags

Same goes for the *F*, feel free to make the diagonal lines parallel.

Is there are special reason for the red jag thing not ending straight at the bottom?

palob:
db entry must be after ``f-spot``, characters like ``-`` are alphabetised before numbers.

maxtron95:
@palob Do you have any suggestions on what to do with the jags on the F1 logo?

palob:
The left inner jag tips should follow a line like this
tips

It occured to me that in order to have the bottom straight you'd have to move the inner tips up by 0.5px. Needs to be seen if this will majorly increase blurriness.
OR have the jags 2px high.

maxtron95:
@palob I am not sure about the bottom jag, I don't think the bottom jag is supposed to be ending straight at the bottom according to the logo.

index

palob:
But the uppermost isn't straight there either. This is basically your jags with the outer right tips moved down by 0.5px.
If (only) one side ends straight it looks a bit askew but in case an icon looking crisp @ 48px can't be achieved otherwise that's the way it is.

maxtron95:
@palob Could I include an icon for F1 2017 in this PR as they could be both related.

palob:
Aye, you can go ahead.

maxtron95:
Added Icons for F1 2017

https://store.steampowered.com/app/515220/

f1-2017

f1-2017-ci
f1-2017-sq

palob (commented):
Did you change the jag distribution for the 2017. I think it works much better for this one. Might be the colour combination.

Foggalong (changes requested):
This translation of the original design needs some work.

Screenshot from 2020-06-06 17-22-42

A key feature of the original design is that those 6 purple highlights are all exactly parallel, not even a tiny deviation amongst them. In the submitted translation at the moment only the two pink highlights are parallel, the three green ones are all slightly off, and the yellow one is _extremely_ off with the point end nodes all being inconsistently places on the x axis.

While we have _some_ wiggle room with this kind of discrepancy in aid of preserving grid alignment I think it should be possible to have these lines parallel in our design too, especially the yellow one which is noticeably wavy even at small sizes.

maxtron95:
@Foggalong I might need help in trying to make this symbol look perfect.

Foggalong:
@creepertron95 sure thing! The way I'd initially approach an icon like this where there's a lot going on is:

1. Take the original image, crop out the symbol and resize to what's wanted (here 30×13px)
2. If there are a lot of parallel diagonal lines, create a template shape with the diagonal wanted
3. Copy that template to all the places a diagonal parallel to that would be wanted
4. Then when creating other shapes, they can clip to that template, rather than the grid points

Screenshot from 2020-07-04 13-17-16

palob (approved):
Your attempt was successful, thank you.

I know this detail is carried over from the original logo but doesn't it look less broken like this?
weird

BlueManCZ:
I actually made a draft of this icon recently and didn't even notice there is a pull request.

a b

palob (approved):

Foggalong (approved):

PR #5130, opened 2020-03-28
Icons for Monster Prom

maxtron95:
https://store.steampowered.com/app/743450/
monster-prom

monster-prom-ci svg
monster-prom-sq

The icons are based on the game's logo.


palob (changes requested):
This time we need a shadow for the arms and/or have them more distinct colour-wise (a bit darker). 
I see the background colour comes from the original game site but I'd prefer something significantly less vibrant.

palob (changes requested):
Thank you. Could you please distribute the nodes and handles like this in order to get a smoother heart?
nodes

palob (approved):

PR #5131, opened 2020-03-28
Icons for Patchwork

palob:
https://github.com/ssbc/patchwork

ssb-patchwork
ssb-patchwork-ci
ssb-patchwork-sq

There is something which feels a bit un-Numixy but I couldn't get it working without the lines.


maxtron95:
@palob It seems that the Papirus Icon theme managed to make an icon for this without strokes. (https://github.com/PapirusDevelopmentTeam/papirus-icon-theme/blob/master/Papirus/64x64/apps/nz.scuttlebutt.Patchwork.svg) 

There should also be a symlink for the icon name, `nz.scuttlebutt.Patchwork`.

palob:
Ah, thanks for the hint. Might try this but we can't have those small in-between shadows.

palob:
This would be my interpretation of this style
ssb-patchwork-ci2

Simplistic approach
ssb-patchwork3

Foggalong:
@palob I love that! Could we have one which uses only 2 or three colours in each section? I think you're right that the detailed interpretation is too detailed but the simplistic one looks a little bare.

Foggalong (changes requested):
@palob I love that! Could we have one which uses only 2 or three colours in each section? I think you're right that the detailed interpretation is too detailed but the simplistic one looks a little bare.

palob:
Pushed
ssb-patchwork-NEW
ssb-patchwork-NEW-sq

Foggalong (approved):

Foggalong:
Amazing work here :heart:

PR #5133, opened 2020-03-28
Icons for Microsoft Teams

### LINKED ISSUE #4971: Icon Request: Microsoft Teams
| --- | :-- | :-- | :-- |
| `Microsoft Teams`| `teams` | `teams.desktop` | original |

Description: `Microsoft Teams is the hub for team collaboration in Office 365 that integrates the people, content, and tools your team needs to be more engaged and effective.`

Website: [`Microsoft Teams`](https://products.office.com/en-us/microsoft-teams/group-chat-software)

### LINKED ISSUE #5159: Update MS Office Icons
As @creepertron95 pointed out [here](https://github.com/numixproject/numix-core/pull/5133#issuecomment-606432270), since we made our MS Office icons the upstream designs have all been overhauled. We're already there Teams but it'd be nice to have the whole suite updated, especially Skype which is Linux compatible.

77995300-2748f900-72f1-11ea-860e-06ccd2261eb3

maxtron95:
Fixes  #4971 

4971-ci svg
sq svg

I also added symlinks for the [teams-for-linux](https://github.com/IsmaelMartinez/teams-for-linux) client


palob (commented):
Could you please remove the ``;paint-order:stroke fill markers;stroke-width:.97895`` stuff from the SVGs?

palob (approved):

Foggalong (changes requested):
I'd sooner see the symbol design for this one make use of the people than the T, though I wouldn't be opposed to a good design which used both.

maxtron95:
@Foggalong I based this off of the other Microsoft office icons to match them.

1*W50NwOOybw55ePEwe97zpg

Foggalong:
@creepertron95 The design currently in Numix are actually based on the previous iteration of Microsoft's icons (maybe even the iteration before?) so the entire set actually needs updating to reflect the redesign. I've opened #5159 to track that :+1:

Foggalong:
Merged into an ms-icons branch for me to work on the redesign today; we've just had to start using the MS Office suite at work :unamused:

palob:
I had to use this in the past too, instead of better suited non-office suite programs. What baffled me most is that they would use LibreOffice whenever table or text stuff had to be imported and saved as xslx, because MS's string handling is severely broken.
There were old MS Office files which would crash MS Office but open fine with LO.

PR #5137, opened 2020-03-28
Icons for Distance

maxtron95:
https://store.steampowered.com/app/233610/

distance-ci
distance-sq


palob (changes requested):
Symbol should go 1px up for Circle & Circle file needs to be optimised.

palob (approved):

PR #5139, opened 2020-03-29
Icons for subtivals

palob:
https://github.com/Subtivals/subtivals

subtivals
subtivals-ci
subtivals-sq


Foggalong (changes requested):
That lighter outline looks a bit off for me. Maybe replacing the fingers/knuckles with separate darker or lighter overlaid shapes might work better?

palob:
Not sure if this works well.
subtivals-2

Foggalong (changes requested):
Oh god no you're right it doesn't. Maybe something involving a projector metaphor instead?

palob:
No way....
subtivals2
Definitely needs a fresh metaphor.

PR #5145, opened 2020-03-30
Icons for Armitage

maxtron95:
http://www.fastandeasyhacking.com/

armitage

armitage-ci svg
armitage-sq svg


palob (changes requested):
That's a spectacular design, thanks for your work. It's a bit detailed (I think for a game icon we can make an exception) and big however. Could we reduce the symbol area by moving the lower hair path 2px up and the avocados? 1px down like this?
Bildschirmfoto vom 2020-03-30 13-28-08
Neck could be shortened by 1px then as well.

Can you please replace ``color-rendering:auto;color:#000000;dominant-baseline:auto;fill:#02c21c;font-feature-settings:normal;font-variant-alternates:normal;font-variant-caps:normal;font-variant-ligatures:normal;font-variant-numeric:normal;font-variant-position:normal;image-rendering:auto;isolation:auto;mix-blend-mode:normal;paint-order:stroke fill markers;shape-padding:0;shape-rendering:auto;solid-color:#000000;text-decoration-color:#000000;text-decoration-line:none;text-decoration-style:solid;text-indent:0;text-orientation:mixed;text-transform:none;white-space:normal`` by ``fill:#02c21c``?

maxtron95:
@palob Actually, this is not a game. It is a frontend for Metasploit. (penetration testing)

palob:
Alright, fancy icon then :)

palob (changes requested):
Thank you for the changes and sorry for not niticing it earlier: this time we've got a bothering outline. Could you please transform the black path to something like this (made the teeth transparent)?
Bildschirmfoto vom 2020-03-30 13-54-20

palob (approved):
Thank you.

Foggalong (changes requested):
I'm gonna reject this one. As @palob says were this a game I'd be tempted to let it through because you've translated it really well, but this is just too detailed for a normal application. I think some variation on the A in the name would work pretty well?

Screenshot from 2020-06-07 18-02-02

maxtron95:
Well that's a shame. However, I redesigned the icon based on the A in the logo.

armitage-ci
armitage-sq

palob (approved):

Foggalong (approved):

Foggalong:
@creepertron95 Thanks for reviewing this, the pushed icon is much more Numixy! I've also added a bit to the design guidelines about having too much icon detail; we had mention there of the the "too little detail" case but not the reverse.

PR #5146, opened 2020-03-30
Icons for AI Dungeon

maxtron95:
https://aidungeon.io/

aid-ci
aid-sq


palob (changes requested):
This has the issue of outlines due to non-overlapping paths. In this case though I think this very effect actually looks good and doesn't need to fixed.

palob:
Ah, sorry for talking BS about what the db needs to have done.

palob:
After ``symlinks`` there needs to be a ``"``.

palob (approved):

PR #5149, opened 2020-03-30
Android Symlinks

maxtron95:
This pull request is for adding android app symlinks to existing icons. 

I will add more symlinks for android apps as I come across them.


maxtron95:
If you ever get here, don't merge this yet as I am looking at all of the apps for android related symlinks. You can fix the tabbing though.

palob:
Maybe this can help in case you don't know: most editors have a setting that when you press *tab* this will be implemented as 4 (or 8 or whatever) spaces.

palob:
What was the way to get the icon names again? Maybe I could add some.

maxtron95:
@palob In the numix android repo, There should be a file, appfilter.xml that lists all of the necessary ComponentInfo.

I am also grabbing apks from the play store and reading through them with ClassyShark for each application's main activity

palob:
Dang!

palob:
Even more reordering needed...

palob:
So is it ``<manifest-->package>/<activity-->parentActivityName>``?
What about ``<application-->name>``?

maxtron95:
@palob If you are reading from appfilter.xml file, Look for a `<item component="ComponentInfo{appid/Activity}"` line for example.

If you are reading the android manifest files. for example Tachiyomi(lines 30-40) [of this file](https://github.com/inorichi/tachiyomi/blob/master/app/src/main/AndroidManifest.xml#L30)

1. look for package id, The package id for Tachiyomi would be `eu.kanade.tachiyomi` found in the line, `package=`

2. look for app activity, the activity would be `.ui.main.MainActivity` found in `android:name=`.
You are looking for activities that have intent filters of `android.intent.action.MAIN` and `android.intent.category.LAUNCHER`

3. the final line you put in data.json is `eu.kanade.tachiyomi/eu.kanade.tachiyomi.ui.main.MainActivity`

maxtron95:
I will go in and add more android application entries.

palob:
What I posted is the variable names shown in classyshark. Can you confirm those are the correct ones?

Ah, got it, maybe. Every package has got a manifest file. It appears the exact syntax differs from app to app. Makes it even more complicated

maxtron95:
@palob What apk are you trying to analyze with Classyshark?

palob:
FDroid.apk
https://f-droid.org/

palob:
Thank you, will try to figure out something.

palob:
``.`` has to be before ``/``, interesting.

palob:
While going through APK for symlinking I realised this:
A disadvantage of beefing up the Android theme is that whenever we want to redesign an icon or change symlinking we'll need to keep track of the Android stuff too.
The combination of Linux and Android in one repo might become a huge maintenance overhead.

We could avoid this if we did only Android "symlinks" in case of 100% match, i.e. if there's respective Linux and Android apps - and we did not "categorical" symlinking as we do for the Linux theme.

However without "categorical" "symlinks" the Android theme will not make big progress.

maxtron95:
@palob Do you think I could maintain the android related stuff? 

I think redesigning an icon will update the android icon as well since the linux and android symlinks are both connected.

The good news is that android app ids rarely change. According to the official [Android Developers page](https://developer.android.com/studio/build/application-id). 

`If you want to upload a new version of your app, the application ID (and the certificate you sign it with) must be the same as the original APK—if you change the application ID, Google Play Store treats the APK as a completely different app. So once you publish your app, you should never change the application ID.`

palob:
The problem I see is that whenever we want to redesign/relink an icon we'll need to check if it still fits to the symlinked (Android) apps.
I'm not sure anymore if it is prudent to built both themes from one core repo or if it is better to duplicate the repo, have branches or whatever.

Let's see what's @Foggalong's musings about an Android reactivation. An initiative of yours will be appreciated for sure, @creepertron95 

I think for now we can go on with this PR.

maxtron95:
Well If we want to reactivate the Android Themes, We should at least bring back the Numix Circle icon theme to the Google Play Store. 

And we need to improve that Icon Request app for Numix. 

Or just do a rewrite of the android themes in [CandyBar](https://github.com/zixpo/candybar-sample).

maxtron95:
@Foggalong What do you think about bringing back the android icon themes again?

Foggalong:
I'm totally fine with having them rebooted! I really miss having complete Numix themes on my phone. That does come with a whole bunch of caveats though. 

Development wise I don't have much input since it was never really my wheelhouse, but my preference would be to use something like CandyBar to make both development and maintenance easier. If you want a run down of how our previous implementation worked, or how icon theme development work(ed) more generally, our old lead dev @paolorotolo might be willing to do that.

I'd sooner be confident that we can actually deliver an app to the appstore again before making a concious effort to bulk up our Android app support, but I'm happy to take PRs for sure especially if they're symlinks to existing icons or otherwise icons which might be useful on desktop too.

@creepertron95 What's your experience with Android development / do you know an Android dev who'd be interested? I've personally done any and I'd sooner spend my own time (at least for now) on making sure the desktop icons are up to scratch instead, so this isn't something I'd be able to take point on beyond what I already do design wise.

maxtron95:
@Foggalong I have basic knowledge of how android icons work. I have used APKTool to decompile and look at the source code of other icon packs. I also manually grabbed apks off Google Play and used ClassyShark to manually look at the Activityinfo. 

We can update the icon request section of our wiki and recommend people to use [Turtl](https://f-droid.org/packages/org.xphnx.iconsubmit/) or [ApplicationsInfo](https://f-droid.org/packages/com.majeur.applicationsinfo/) for Icon requests, similar to what @dkanada's [Frost icons](https://github.com/dkanada/frost) suggests on their description. 

Regarding the App Store, we can try to give a shot at publishing to F-Droid as well.

I would be interested to know how icon development used work at Numix just to see how you guys did it.

I feel that using CandyBar would be a good idea to simplify things.

P.S. I already created a few PRs for Android icons in the Numix Core repo, I marked those icons (Android) in the title.

Foggalong:
@creepertron95 Excellent! In which case more than happy for you to get the ball rolling on this. First step is probably just to put together a proof of concept, modifying CandyBar to use Numix branding and a couple of Numix icons. I've dropped Paolo a message to see if he'd be up for that chat about the old versions too.

Once that's sorted we can start working on updating the icon request information and brining that up to date, then start working on building a more complete version to push to the Play Store. We did have versions in F-Droid but they've been dropped since we stopped regularly updating them, so that's something we can restart around then too (though perhaps counter intuitively getting an app in F-Droid is more involved than getting it in Play).

data.json
```diff
@@ -13,6 +13,60 @@
...
+            "com.v5d2048.mobidev/com.d2048.mobidev.MainActivity",
+            "com.yotadevices.yotaphone2.game2048/com.yotadevices.yotaphone2.game2048.MainActivity",
+            "fun2048.fungame.win/com.d2048.mobidev.MainActivity",
+            "game2048.b2048game.twozerofoureight2048.game/game2048.b2048game.twozerofoureight2048.game.MainActivity",
+            "game2048.b2048game.twozerofoureight2048.game/game2048.b2048game.twozerofoureight2048.game.SplashActivity",
+            "game2048.c2048game.twentyfortyeight.twozerofoureight2048.game/game2048.c2048game.twentyfortyeight.twozerofoureight2048.game.C2048Game",
+            "in.net.shreyas.game2048/in.net.shreyas.game2048.Start",
+            "it.zerocool.game2048/it.zerocool.game2048.MainActivity",
+            "org.mania2048/org.mania2048.NumberPuzzle"
+        ],
```
Foggalong, line 69: jfc everyone and their mum really did make a 2048 clone didn't they

Foggalong (approved):

data.json
maxtron95: If you thought the 2048 clones were bad. Just wait until you see all of the Flappy Bird clones on Google Play.

paolorotolo:
Hey guys! Sorry for the late reply. I'm really excited to revamp Android themes again, I think we might open another issue or discuss somewhere else to plan something about it. I can give an hand. Yeah it would be easier to use candybar, but I'm thinking about CI to automate building the different icon packs. Plus, when we introduced the first version of the icon packs, Android didn't support shapes for icons. Now the situation is a bit different, we might be able to ship adaptive icons (circle, square, squircle...) with a single icon pack integrating with the system, but I have to read more about how it works.

Foggalong:
Continuing the conversation over in the Android Circle repo, seemed like the most sensible place!

PR #5152, opened 2020-03-30
Icons for Baldi's Basics

maxtron95:
https://www.basicallygames.com/baldis-basics.html

baldis

bb-sq
bb-ci

Alternative Baseplate Color:
bb-alt


palob:
Maybe a (fair way) lighter variant of the alternative colour? As is it lacks some contrast.

maxtron95:
I pushed the alternative color baseplate but brighter for the symbol to contrast with it.

palob:
I think it could be a bit brighter and less saturated still.

palob (changes requested):
There's some SVG cruft left which I'd like you to ask to remove.

Baseplate is alright, could be ever so slightly brighter.

palob (approved):

PR #5158, opened 2020-03-31
Icons for 1Weather (Android)

maxtron95:
https://play.google.com/store/apps/details?id=com.handmark.expressweather

(1Weather)com handmark expressweather

1weather svg
1weather-sq svg


palob (changes requested):
This is now perfectionism but the roundings look a bit awry. The symbol should be built with rounded rectangles or ellipsoids.

awry

Same for the white inner body (the rounded corner on the left will be a bit tricky, you have to add a triangle minus circle), rectangles/circles (and a stroke probably).

The contour line looks a bit harsh against the wight cloud body. Could we get it a bit brighter and less saturated, more like in the bottom right region of the original icon?

db entry should be after ``140-game``.

palob (approved):
Oh yeah, thanks.

PR #5169, opened 2020-04-02
Icons for Dradis

maxtron95:
https://github.com/dradis/dradis-ce

dradis

dradis-ci
dradis-sq


palob (changes requested):
Could you get the symbol a bit smaller without going out of your way?
The inner surface of the bottom leg of the **A** should actually be the brighter one with the Numix source of light positioned above the icons.

The lower end of the bottom leg should look more like a square, you're free to move nodes to .5px osition for that.

Foggalong (changes requested):
I'm a bit wary about making this design 3D, not least because it's going to mean really carefully picking what faces need to be what colour of green to accurately reflect the lighting we use. Is there a way this can be flattened?

palob (changes requested):
Very simplistic but works. Both symbols should sit 1px higher though.

Foggalong (commented):
Much happier with this design, will merge once the symbol has been raised :+1:

palob (approved):

Foggalong (approved):

PR #5171, opened 2020-04-03
Icons for Mimo (Android)

maxtron95:
https://play.google.com/store/apps/details?id=com.getmimo

com getmimo

mimo-ci
mimo-sq

Alternative Icon (Based on official iOS icon):
mimo-alt


palob (changes requested):
I know it's not technically a doble shadow, it just looks like one but can you alter it (merge & intersect) so that it looks like this (cf. original icon)?

Could you please smoothen the cornery nodes of those 2 paths?


There's some SVG cruft again.

maxtron95:
@palob The icon should be fixed now.

palob (commented):
Thanks for the changes, just if you have got some time to spare, could you please fix alignment of the mouth path as seen in the screenshot?

palob (approved):

Issue #3706, opened 2017-04-30
Jetbrains icons look out of place


heyhippari:
The redisigned Jetbrains icons have a wild gradient that goes way out of the 10L guideline.

AFAIK, they're the only icons to do that and, as a result, look quite out of place.

I suggest averaging the two colors of the current gradient and then applying a 10L gradient as usual.

Examples:
![idea-old](https://cloud.githubusercontent.com/assets/19396809/25564418/3eb26a1c-2db3-11e7-9a33-25952f295868.png) => ![idea-new](https://cloud.githubusercontent.com/assets/19396809/25564420/44045912-2db3-11e7-9ada-c3dc0ee136c2.png)

palob:
When we merged those we intentionally opted for a 2-colour-gradient as there was no single characteristic colur in the oiginal icons. There are a few more icons with a 2-colour gradient, for instance curlew and soundnode.

Foggalong:
Ideally I would like to get rid of dual gradients in all icons, so I'm gonna leave this open. If anyone does have any good ideas how to make this happen so that it works across all the dual gradient icons I'd love to hear them.

@MrTimscampi I'd be interested to know if the averaging works for all the Jetbrains ones

heyhippari:
@Foggalong I haven't tested all of them, but PyCharm and IntelliJ work pretty well. Stuff that has wildly different colors might be less perfect, though.

I'm busy until early July with finals and internship/thesis presentations, but I'll try to make a few of them if/when I have time, if you want.

heyhippari:
Update on this, here are some tests for the most jarring ones I found.

**Rubymine**:
![rubymine old](https://user-images.githubusercontent.com/19396809/29824510-0ba28fb6-8cd2-11e7-867f-399237b2977c.png) -> ![rubymine](https://user-images.githubusercontent.com/19396809/29824189-14618cfc-8cd1-11e7-8fa2-c0137b26e671.png)

**Rider**:
![rider old](https://user-images.githubusercontent.com/19396809/29824522-13762b80-8cd2-11e7-9ec6-623c4389ecfd.png) -> ![rider](https://user-images.githubusercontent.com/19396809/29824529-172bed28-8cd2-11e7-9ca7-bd13d4c78f81.png)

If the 10L guideline is still applicable (I saw that the guidelines changes a bit while I was away :p ) and those look alright, I can make a PR with everything.

Foggalong:
[...]

PR #5172, opened 2020-04-03
Icons for Happy Wheels

maxtron95:
https://play.google.com/store/apps/details?id=com.fancyforce.happywheels

com fancyforce happywheels

happy-wheels-ci
happy-wheels-sq


palob (changes requested):
Symbol is very nice but two-colour-baseplates is another design feature set to be retired, cf. #3706 for instance.
Again some style cruft and Circle icon isn't saved as optimised.

maxtron95:
@palob Should I go ahead and average the colors for the baseplate?

palob:
average, one out of the 2 or a new colour, whatever will cut a good figure and/or have some relation to the app.

palob (approved):

Foggalong (changes requested):
This is just too detailed unfortunately and doesn't end up fitting in the Numix style. A bicycle or the white helmet (without a head) viewed side on might give some more joy designwise

palob (changes requested):
Could we please have the alignment fixed like this



and the helmet colours set to ``f9f9f9ff`` (white) and ``333333ff`` (black)?

palob (approved):

PR #5174, opened 2020-04-04
Icons for TikTok

maxtron95:
https://www.tiktok.com/en/

tiktok

tiktok-ci
tiktok-sq


palob (changes requested):
I'm afraid this doesn't work. The baseplate needs to be brighter as the symbol shadow will be hard to recognise this way.
A gradient of 51L --> 61L is the darkest we normally use for blacl/grey baseplates.

The symbol shadow ought to be black with 10% opacity, sorry and we cant have this on-symbol outline.

maxtron95:
@palob How do you think I should go about the symbol then? Many people recognize the tiktok icon from the blue and red outlines present on the original icon.

palob:
I think it's a bad design even with the original icon. Looks glitched at small sizes. Maybe blue for the symbol and the red for the baseplate?

maxtron95:
@palob I will see if I can come up with some alternative designs after I work.

maxtron95:
@palob I created some alternative icons. You can pick which one is the best and I will go from there.

1.
tiktok-alt1

2.
tiktok-alt12

3.
tiktok-alt3

palob:
I say no 1.

palob:
Does the symbol on the Circle icon more centered when shifted 1px to the left and/or 1px up?

Foggalong (commented):
Staring at these I think maybe up 1px but not left?

Screenshot from 2020-06-28 17-36-06

palob (approved):

PR #5175, opened 2020-04-04
Icons for Snapchat (Android)

maxtron95:
https://www.snapchat.com/

snapchat

snapchat-ci
snapchat-sq


palob (changes requested):
I see it's from the originl icon but this a very garish colour combination for Numix.
Could you try a brighter and less saturated variant of the baseplate?

These lines should begin straight as in the original icon.
ttt

palob (changes requested):
Baseplate is better now but yellow+white still is a combination I'm not particularly fond of.

Thanks for the straightened sides. However the transition from straight to curve should be smooth.

curving

Foggalong:
I was under the impression we already had a snapchat icon! Will go digging for it

Foggalong:
@creepertron95 It would seem we have a lot of Android icons which were in those themes but never ported over to the Core system: [Circle](https://github.com/numixproject/com.numix.icons_circle/tree/master/Icons) and [Square](https://github.com/numixproject/com.numix.icons_square/tree/master/Icons). I was under the impression I had merged them across but apparently not! It'll be worth one of us digging through that folder and seeing what's worth salvaging from there.

palob (approved):
I still don't like the colour combination and the outline from the original icon but I can't think of an obvious workaround.

PR #5177, opened 2020-04-06
Icons for MediBang Paint

maxtron95:
https://medibangpaint.com/en/

medibang-paint

medibang-paint-ci
medibang-paint-sq


palob (changes requested):
Circle symbol should be shifted 1px to the left+ 1px up, for Square maybe 1px up.

In my opinion the green and cyan parts don't contrast well against the baseplate. Could you try a "black" (L51 --> L61) one?

palob (approved):

PR #5178, opened 2020-04-06
Icons for Miminote (Android)

maxtron95:
https://play.google.com/store/apps/details?id=com.moquji.miminote

miminote

miminote-ci
miminote-sq


palob (changes requested):
Could you please fix the dog-ear shadow on the Square symbol so that it looks like this?
dogear

palob (approved):

PR #5179, opened 2020-04-06
Icons for MALClient (Android)

maxtron95:
https://play.google.com/store/apps/details?id=com.drutol.malclient

com drutol malclient

malclient-ci
malclient-sq


palob (changes requested):
Could just use a small simplification in this area
simple

palob (changes requested):
OK, this is very minor but there are still some ever so slight misalignments, fixed in the screenshot below.
small

You should run Path -> Union on the symbol path in order to get rid of a inner intersection.

palob (approved):

PR #5188, opened 2020-04-08
Icons for Doodle Jump (Android)

maxtron95:
https://play.google.com/store/apps/details?id=com.lima.doodlejump

doodle-jump

doodle-jump-ci
doodle-jump-sq

### DIFF
--- a/data.json
+++ b/data.json
@@ -5163,6 +5163,11 @@
             "root": "donut"
         }
     },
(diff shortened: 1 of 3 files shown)


palob (changes requested):
Lovely design, circle icon should go 2px to the right, square 1px.

palob (approved):

PR #5189, opened 2020-04-08
Icons for Candies 'n Curses (Android)

maxtron95:
https://play.google.com/store/apps/details?id=com.tomtominc.candiesncurses

com tomtominc candiesncurses

cnc-ci
cnc-sq


palob (changes requested):
Design is fine visually but technically there are many cases of strange stacking, unneeded paths and nodes. Should be like this for instance.
Bildschirmfoto vom 2020-04-08 10-19-18

There's two cruft blocks in the files? Could you please cut down to ``fill`` or ``opacity``?

Foggalong (commented):

PR #5195, opened 2020-04-09
Icons for preferences-web-browser-cache

maxtron95:
For #2294

128px-Oxygen480-apps-preferences-web-browser-cache svg

cache-ci
cache-sq


palob (approved):
It's a bit high but I think it works. Will be no biggie to get the height down to 28px otherwise.

PR #5196, opened 2020-04-09
Icons for preferences-web-browser-identification

maxtron95:
For #2294 

128px-Oxygen480-apps-preferences-web-browser-identification svg

browser-id-ci
browser-id-sq


palob (commented):
Now that we may want to get rid of the full-baseplate globe web browser icons the globe could become a misrepresentation in future.
So perhaps just without it or a rectangle representing a photo?

palob (commented):
Thank you for the changes.
I think it'll look cleaner with a consistent 2px margin.
margin

Foggalong (changes requested):
Agreed; looks good now but will be better with that nice margin :+1:

maxtron95:
@palob Do you mean margins as in the black bars on the bottom of the card?

palob:
There should be a 2px padding, meaning any element should keep it 2px distance. Black & blue bars, globe. (For the what is shown above I reduced the globe size)

palob (approved):

Foggalong (approved):

PR #5198, opened 2020-04-13
Icons for veloren

palob:
https://gitlab.com/veloren/veloren

net veloren veloren
veloren-ci
veloren-sq

### DIFF
--- a/data.json
+++ b/data.json
@@ -19836,6 +19836,14 @@
             "root": "vega"
         }
     },
+    "veloren": {
+        "linux": {
+            "root": "veloren",
+            "symlinks": [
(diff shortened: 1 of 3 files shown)


Foggalong (changes requested):
Is there a way to do this without the black outline?

palob:
With the outline turned white it looks damn pixelated.
veloren-ci2

Foggalong:
Oof yeah no that's not good. I will merge this but I think at some point we need to have a think about how we best integrate pixelated icons into the design of the theme.

Foggalong (approved):

PR #5204, opened 2020-04-14
Icons for Broforce 

### LINKED ISSUE #2825: Icon for Broforce (Steam)
Game: Broforce

AppID: 274190

Icon: [http://www.blackshipsfillthesky.com/BROFORCE/wall2.jpg](url)

Notes: Every action movie you've every watched in a single game e.g. Rambo, Blade, Matrix, Die Hard etc..

maxtron95:
Fixes #2825 

broforce

broforce-ci
broforce-sq


palob (changes requested):
Could you please try brighter colours? Right now it looks like viewed through sunglasses.

palob (changes requested):
This is much better but still not a colour combination which will make me shout out with joy.
Especially the baseplate feels off. Where did you take the colours from? Maybe there is an alternative colour scheme?

The paths composing the star share the same outline which causes rendering glitches.

maxtron95:
The colors are from the original game's logo.

Foggalong (changes requested):
This is a _very_ 3D design; if I'm reading it right, the orange of the start at the bottom is to signify a light shining from below? Either way the design is way out of place in Numix because of the lighting and 3D perspective. Would it look okay with the star flattened out and centred?

Issue #4854, opened 2019-04-16
Icon Request: List of Steam Games


KRACK-BIT:
| game | SteamID | Original Icon | Notes |
| --- | :-- | :-- | :-- |
| XCOM 2 | 268500 | ![image](https://user-images.githubusercontent.com/46194176/56198182-51e36880-6032-11e9-9d41-9cbecdac2367.png) | the predecessor, XCOM: Enemy Unknown has a numix icon |
| Keep Talking and nobody explodes | 341800 | ![UnityPlayer](https://user-images.githubusercontent.com/46194176/56206868-451c4000-6045-11e9-8429-b183e5edd590.png) | team game about defusing a bomb |
| Turmoil | 361280 | ![icon](https://user-images.githubusercontent.com/46194176/56206813-274edb00-6045-11e9-9acc-5eb783d07b46.png) | simulation game of being an oil prospector; very visually charming |
| Train Valley | 353640 | ![image](https://user-images.githubusercontent.com/46194176/56206981-890f4500-6045-11e9-871d-f1b94f2daf03.png) | train puzzle game |
| Train Valley 2 | 602320 | ![image](https://user-images.githubusercontent.com/46194176/56207111-d986a280-6045-11e9-8c10-3ce7bff9eeb8.png) | sequel to the prev. game; same genre |
| The Talos Principle | 257510 | ![image](https://user-images.githubusercontent.com/46194176/56207401-86f9b600-6046-11e9-819b-a168011b2c9b.png) | Well-acclaimed first-person Puzzle Game |
| SquareCells | 416770 | ![image](https://user-images.githubusercontent.com/46194176/56207463-ae508300-6046-11e9-8f23-52b2279d2ac1.png) | ambient logic puzzle game |
| SlayAway Camp | 530390 | ![UnityPlayer](https://user-images.githubusercontent.com/46194176/56207569-fb345980-6046-11e9-8946-6995caad1056.png) | Cartoon killer puzzle game |
| Polybridge | 367450 | ![image](https://user-images.githubusercontent.com/46194176/56207666-38005080-6047-11e9-9f43-88222b6ad7d7.png) | physics-based puzzle solver |
| Planetary Anihilation: TITANS | 386070 | ![image](https://user-images.githubusercontent.com/46194176/56207771-831a6380-6047-11e9-936b-b38b3d2676e8.png) | interstellar battle strategy game |
[...]

PR #5205, opened 2020-04-14
Icons for Clustertruck

maxtron95:
For #4854

ctruck

clustertruck-ci
clustertruck-sq


palob (approved):

Foggalong (changes requested):
How would this one look without the outlines around each shape? We don't usually do too many outlines and I have a feeling this would work without them (though I might be wrong, in which case I'd approve this design)

maxtron95:
@palob @Foggalong Something like this?

clustertruck-new svg

palob:
Yes :+1: Plus maybe an on-symbol shadow for the front cabin.

palob:
@creepertron95 Do you happen to have the latest draft somewhere on your disk?

PR #5209, opened 2020-04-15
Icons for Pixel Vision 8

maxtron95:
https://www.pixelvision8.com/

pixelvision8

pv8-ci
pv8-sq

### DIFF
--- a/data.json
+++ b/data.json
@@ -14376,6 +14376,11 @@
             ]
         }
     },
+    "pixel-vision-8": {
+        "linux": {
+            "root": "pixel-vision-8"
+        }
+    },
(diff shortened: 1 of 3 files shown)


palob (changes requested):
Square symbol should go 1px upwards.

maxtron95:
But If I do that, the pixel logo will end up being blurry at 24px.

palob:
Hmm, @Foggalong what do you think is to be prioritised?

Foggalong:
Stick it 1px up for now - icons optimised for smaller sizes is something to return to at a later date

palob (approved):

PR #5212, opened 2020-04-15
Icons for A1 Telekom (Android) 

maxtron95:
https://www.a1.group/en/home

a1-telekom

a1-telekom-ci
a1-telekom-sq


palob (changes requested):
I think we should ditch the 3Dness here.

Foggalong (commented):

PR #5213, opened 2020-04-15
Icons for 94% (Android)

maxtron95:
https://play.google.com/store/apps/details?id=com.scimob.ninetyfour.percent

94percent

94p-ci
94p-sq


palob (changes requested):
There's seems to be something wrong with the **9**. I guess it should have been a full 1px ring there.

palob (changes requested):
It's still there. I was referring to the misalignment at the top of the **9**.

Foggalong (commented):

PR #5221, opened 2020-04-20
Icons for Party Hard

maxtron95:
For #4854 

phard-ci
phard-sq


palob (changes requested):
Symbols can go 1px to the right.

palob (approved):

Foggalong (changes requested):
Is there a way to adjust / re-position the lower red slash so that it looks less like a grin?

PR #5222, opened 2020-04-23
Icons for Planetary Anihilation: TITANS 

maxtron95:
For #4854 

pat-ci
pat-sq


palob (changes requested):
Symbol on Circle should sit 1px further down.

palob (approved):

PR #5223, opened 2020-04-23
Icons for Poly Bridge

maxtron95:
For #4854 

poly-bridge-ci
poly-bridge-sq


palob (approved):

Foggalong (changes requested):
Like the idea, but if being viewed head on the camper needs to be wider. The design in the game and in the icon is indeed very thin, but almost always viewed side on.
Screenshot_4-694x393

The campervans they use in promotional materials that are seen head on are often adjusted so they're not as thin looking.
Header_post-820x353
poly-bridge-switch-hero

And actual VW Campervans are even wider still (though I don't think we have to go that wide.
airfix-j6024-quickbuild-vw-camper-van-bleu

Requested change would either be to make the current design much wider by several pixels in both directions, or if you prefer to switch the design so the van is viewed side-on. I don't have a preference for which at this point so I'll leave that to you :)

PR #5224, opened 2020-04-23
Icons for Slayaway Camp

maxtron95:
For #4854 

slayaway-camp-ci
slayaway-camp-sq


palob (commented):
I think it's OK as is but how about removing that outer rim which seems to be the background in the original icon? Good idea?
alt

maxtron95:
I'm not sure about that seeing the character from the the game icon

skillface

palob:
Ah I see. Is it possible to get the whole character represented or will this make the skull too fine-grained?

Foggalong:
@palob Took the liberty to push the start of a fix for that issue; make the head the colour of the flesh. There's some more to do like playing with adding the straps and merging the flesh-coloured shapes to be one, but I'll leave finishing that to @creepertron95.

palob (approved):

PR #5225, opened 2020-04-23
Icons for SquareCells

maxtron95:
For #4854 

squarecells-ci
squarecells-sq


palob (changes requested):
I'd prefer a less garish, better contrasting combination of colours. More like the original icon. Or darker baseplate + brighter symbol or bright but less saturated baseplate + darker symbol.

PR #5234, opened 2020-04-25
Icons for Turmoil

maxtron95:
For #4854 

turmoil-ci
turmoil-sq


palob (changes requested):
I think for the **X** structures you can use narrower strokes. The drill rod should be darker like in the original icon and the tower structure should shadow on it.

Background could both be a bit lighter like in some areas of the original icons.

PR #5236, opened 2020-04-25
Icons for A Hat in Time

maxtron95:
https://store.steampowered.com/app/253230/

a-hat-in-time

a-hat-in-time-ci
a-hat-in-time-sq


palob (changes requested):
Needs something to counter the outline issue.
nodes

With the light source at the top left corner (Circle) or at the center top (Square) that shadow shape should actually be on the on the right side of the hat or be removed altogether.

palob (changes requested):
Forgive my perfectionism, there's some small strange jags
perfect
.

PR #5239, opened 2020-04-25
Icons for VA-11 Hall-A

maxtron95:
https://store.steampowered.com/app/447530/

va-11-hall-a-ci
va-11-hall-a-sq

### DIFF
--- a/data.json
+++ b/data.json
@@ -19351,6 +19351,11 @@
             "root": "v4l2ucp"
         }
     },
+    "va-11-hall-a": {
+        "linux": {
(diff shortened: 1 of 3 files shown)


palob (changes requested):
Lovely design idea! I think the bars could be distributed evenly like this
bars

palob (changes requested):
Could you please shift the bars down 1px to reduce the clash with the foam head?

palob (approved):
That's ace, thank you.

PR #5241, opened 2020-04-25
Icons for OneShot

maxtron95:
https://store.steampowered.com/app/420530/

oneshot-ci
oneshot-sq


palob (changes requested):
On both icons the symbol should sit 1px lower.
Could you please tone down the white parts to ``ededcbff``. Right now the contrast is a bit stark.

Now this is a bit nitpicky, could you please straighten the wire like this (by merging with 0.5 x 1 rectangles)?
wendel

palob (changes requested):
Thanks but could you please striaghten the wire on both sides? Sorry for my ambiguous illustration. Coulour seems to be a bit too saturated now + the sparks should be the same colour.

PR #5247, opened 2020-04-26
Icons for osu!

maxtron95:
https://osu.ppy.sh

osu

osu-ci
osu-sq

Note: There is a linux version but, it is appimage only.


palob (commented):
Could you get this closer to the original? The font looks much like [Ubuntu](https://fonts.google.com/specimen/Ubuntu) probably preinstalled on your system

palob (approved):

Foggalong (changes requested):
@palob I think you accidentally approved?

But also as my actual review, regardless of what font change a lighter pink background behind where the text cut-out is would be nice

palob:
Looking on it again, whatever happens to the font, I think it should be less close to the circle limits.

LiquidPL:
If I'm not wrong, [this](https://www.fontsquirrel.com/fonts/aller) is the font used in the original logo.

palob:
Additional icon names 
``sh.ppy.osu``
``sh.ppy.osulazer``

PR #5248, opened 2020-04-26
Icons for Botanicula

maxtron95:
https://store.steampowered.com/app/207690

botanicula

botanicula-ci
botanicula-sq


palob (changes requested):
Both symbols can go 1px up and there's large cruft blocks.

There's a slight avoidable misalignment but maybe we can let this slip in.
misalign

palob (changes requested):
Sorry, I'm now viewing this on a low DPI screen and the misalignment shows as perceivable bluriness. Would you mind straightening up the right side of the symbol?

PR #5251, opened 2020-05-01
Icons for Zero-K

maxtron95:
https://zero-k.info/

zero-k-ci
zero-k-sq


palob (changes requested):
It's difficult to understand what the icon is supposed to show.
Something seems off with colours or proportions or positioning.

palob:
db entry needs one more closing bracelet.

maxtron95:
The icon is supposed to look like a sun and moon.

zerok

palob:
Maybe it's a good idea to move the moon to the upper left and the sun to the bottom right. This way the moon could cast a shadow on the sun enhancing the 3D effect. Actually a shadow is already possible with the simple left-right composition of the original icon.

Don't know if the glow on the moon is achievable without a gradient. For two solid colours I suppose they need to be closer together and closer to the original ones than the ones in your draft.
Maybe for the glow around the moon on the left you could try something as the outer jagged rim in our ``xcaliber.svg`` or the simple rim in ``redeclipse.svg``.

PR #5252, opened 2020-05-01
Icons for Yooka-Laylee

maxtron95:
https://store.steampowered.com/app/360830/

yooka-laylee

yl-ci
yl-sq


Foggalong (changes requested):
The shading here is off; in the original icon the light source is from the top right but in Numix icons it's top left. I'm not sure what adjustment needs making here, maybe just flipping the shades and then making the difference between the two smaller, but it does need a tweak.

PR #5256, opened 2020-05-04
Icons for poi

maxtron95:
https://github.com/poooi/poi

poi

poi-ci
poi-sq


palob (changes requested):
At the bottom the symbol shadow has got quite a few unneeded nodes.

The eyes look a bit creepy. Could you try adding the swirl from the original icon and enlargening the white part?
The hair could stand out a bit more against the skin colour.

I'm not sure if the icon could work better without the circle
better
(symbol position and shadow would have to be adjusted ofc)
The circle adds a nice touch of colour and a window effect but I'm not sure it make the design look busier than needed.

There's some SVG cruft left.

PR #5259, opened 2020-05-05
Icons for GameMaker Studio 2

maxtron95:
https://store.steampowered.com/app/585410/

gms2

gms2-ci
gms2-sq


palob (approved):

palob:
@creepertron95 Could you please render your preview PNGs at an integral multiple of 48px X 48px, for instance 96x96 or 144x144 in future? This way bluriness in the preview icons will tell about misalignments and can't be caused by the scaling.

PR #5261, opened 2020-05-05
Icons for VRChat

maxtron95:
https://store.steampowered.com/app/438100/

vrchat

vrchat-ci
vrchat-sq


palob (changes requested):
Both symbols should be moved 1px downwards and the inner node of the **V** should go 1px up.
V

palob (approved):

PR #5264, opened 2020-05-08
Icons for Crazy Taxi

maxtron95:
https://store.steampowered.com/app/71230/
https://play.google.com/store/apps/details?id=com.sega.CrazyTaxiplayhaven&hl=en_US

crazy-taxi

crazy-taxi-ci
crazy-taxi-sq


palob (changes requested):
Could you please fix alignment here
align
and at the symbol bottom?

The orange-yellow on the symbol should be more different from the baseplate colour or the symbol should be transparent in those spots with according shadow.

PR #5265, opened 2020-05-08
Icons for A Boy and His Blob

maxtron95:
https://store.steampowered.com/app/281200/

a-boy-and-his-blob

abahb-ci
abahb-sq


palob (changes requested):
Circle symbol should be placed at X=11, Y=13

palob (approved):

PR #5268, opened 2020-05-08
Icons for Donut County

maxtron95:
https://store.steampowered.com/app/702670/

donut-county

donut-county-ci
donut-county-sq


palob (changes requested):
Wonderful design! The black ear parts should be reduced to this in order to avoid the outline.
ears

PR #5270, opened 2020-05-08
Icons for Forager

maxtron95:
https://store.steampowered.com/app/751780/

forager

forager-ci
forager-sq


palob (changes requested):
Could you please remove the stroke-path conversion SVG cruft blocks?

palob (approved):
This is good to go if the outline is OK.

Foggalong (changes requested):
Yeah the outline stops this being Numixy. There are a couple of other tweaks I would have suggested like changing the shape of the eyes slightly, making the mouth the open variant, and removing the bottom shadow (it looks weird without the outline) buuuuut....

Foggalong:
... in making an icon to compare it without the outline I ended up getting carried away and making those changes. It's just too cute an icon :heart_eyes: 

 forager-circle forager-square

palob (approved):

maxtron95:
@Foggalong Your icon definitely looks better than mine.

PR #5274, opened 2020-05-09
Icons for Narcissu 1st & 2nd

maxtron95:
https://store.steampowered.com/app/264380/

narcissu

narcissu-ci
narcissu-sq


palob (approved):
Don't know where you got that special **n** design from but I like it.

maxtron95:
The design is actually the first letter of the game's logo.

Foggalong (changes requested):
I'm gonna have to disagree:

1. On the left vertical part of the n, the right hand lines should follow on from each other as in the original.
2. The left and right vertical parts should be the same width as each other

I do like the way you've mounted the n on an extra shape though, makes the icon more interesting!

PR #5275, opened 2020-05-09
Icons for NEKOPARA

maxtron95:
https://store.steampowered.com/app/333600
https://store.steampowered.com/app/385800
https://store.steampowered.com/app/420110
https://store.steampowered.com/app/602520
https://store.steampowered.com/app/899970

nekopara

nekopara-ci
nekopara-sq


palob (changes requested):
I think the paw will be easier recognisable as such with the shapes looking something like this.
Black part could be straightened like in the original icon.

paw

Right now the circles/ellipses are a bit close together.

There's some stroke conversion cruft left.

palob (changes requested):
Could you please adjust the outlining black shape to the modified circles (I didn't do this in the screenshot) and remove the SVG cruft?

PR #5279, opened 2020-05-10
Icons for IPFS Desktop

palob:
https://github.com/ipfs-shipyard/ipfs-desktop

ipfs-desktop
ipfs-desktop-ci
ipfs-desktop-sq


Foggalong (changes requested):
I like this, but it looks like the cube should be solid with faces rather than hollow. I think use the current baseplate colour for the faces (as in the original) and maybe a light turquoise-grey for the baseplate?

palob:
Like this?
ipfs-2

Foggalong:
@palob Aye but with a much lighter background

palob:
Next try
ipfs2

Foggalong:
That's not what I was envisioning but it actually looks better! Good to go :+1:

palob:
Pushed.

Foggalong (approved):

PR #5280, opened 2020-05-10
Icons for NieR:Automata

maxtron95:
https://store.steampowered.com/app/524220/

nier-automata

nier-automata-ci
nier-automata-sq


palob (commented):
Spectacular design, is there an at least halfway decent looking way of reducing the tooth number for the sake of grid alignment?

Foggalong:
@palob Thoughts?

Rounded middle line
automata-192
Straight middle line
automata-192striahgt
No middle line
automata

Foggalong:
Comparing it to @creepertron95's design, the aligned-teeth versions definitely loses out on the creepiness that comes from having so many teeth in the mouth. I'm undecided, what do you reckon?

palob:
I think they all look rather mischievous than creepy. How about jagged teeth, approximating the original teeth with triangles?

Foggalong:
Triangles might be worth a shot, but I'll likely not have time to delve into trying that until the weekend so if anyone else fancies a bash go for it :+1:

palob:
Triangular or half-ellipse teeth of 1px width (like in the original icon) with 1px black space in between might be OK.
Can look into this next week.

PR #5284, opened 2020-05-15
Icons for Vectorman

maxtron95:
https://store.steampowered.com/app/34280/
https://play.google.com/store/apps/details?id=com.sega.vectorman

vectorman

vectorman-ci
vectorman-sq


palob (changes requested):
Spectacular design, could we have the **V** stick out more at the bottom of the ball?

PR #5286, opened 2020-05-15
Icons for Torchlight

maxtron95:
https://store.steampowered.com/app/41500/

torchlight

torchlight-ci
torchlight-sq


palob (approved):

Foggalong (changes requested):
Can we get a shield in the background? I think it'd make this icon a bit more interesting

PR #5288, opened 2020-05-15
Icons for Super Inefficient Golf

maxtron95:
https://store.steampowered.com/app/772480/

super-inefficient-golf

super-inefficient-golf-ci
super-inefficient-golf-sq

### DIFF
--- a/data.json
+++ b/data.json
@@ -17817,6 +17817,11 @@
             ]
         }
     },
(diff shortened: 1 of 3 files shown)


palob (changes requested):
Could we please have this as symbol-on-symbol design, making it more obvious that this is supposed to be a ball?

PR #5292, opened 2020-05-18
Icons for Source Filmmaker

maxtron95:
https://store.steampowered.com/app/1840/

sfm

sfm-ci
sfm-sq


palob (changes requested):
Could you please try a radius of 4px for the outer circles and a radius of 2px for the inner circle?
Perhaps this way the upper circle could be placed at integer coordinates.

PR #5293, opened 2020-05-18
Icons for Samorost 3

maxtron95:
https://store.steampowered.com/app/421120/

samorost-3

samorost-3-ci
samorost-3-sq


palob (changes requested):
Could you please try a bigger head with the head sticking out in a bigger area + bigger eyes?

PR #5295, opened 2020-05-20
Icons for Reassembly

maxtron95:
https://store.steampowered.com/app/329130/

reassembly-ci
reassembly-sq


palob (changes requested):
Circle symbol should be 1px lower.

palob (approved):

PR #5297, opened 2020-05-21
Icons for Geonkick

palob:
https://gitlab.com/iurie-sw/geonkick

geonkick
geonkick-ca
geonkick-ci


Foggalong (approved):
I was pleasantly surprised how well this scales to smaller sizes for such a detailed icon! Excellent design work

PR #5307, opened 2020-05-22
Icons for Psychonauts

maxtron95:
https://store.steampowered.com/app/3830/

psychonauts

pnauts-ci
pnauts-sq


palob (commented):
Does the spiral design work when shrinked to 30x30?
Symbol should be no larger than 22x28.

palob (changes requested):
Size is OK now but I think the spiral makes it look quite busy? How does it look with (aligned) rings instead or without anything behind the character at all?

I'm also not sold on the colours. Could you please try a darker colour for the character and an orange background like in the original icon?

Foggalong (commented):

PR #5311, opened 2020-05-27
Icons for Portal Stories: Mel

### LINKED ISSUE #3170: Icon for Portal Stories: Mel (Steam)
Game: Portal Stories: Mel
AppID: 317400
Current icon: 
portal2

Notes: It _is_ a Portal 2 mod, however it has its own launcher/game data so I believe that shouldn't be an issue.

maxtron95:
Fixes #3170

portal-stories-mel-ci
portal-stories-mel-sq


palob (approved):

Foggalong (changes requested):
The spacing between the pinwheeling blue segments needs to be consistent rather than variable (parallel lines again). This _looks_ like it should be a minor tweak but is actually really complicated to set up with guidelines and using Inkscape's rotation tool. I'm going to make this change myself, but for future reference these are shapes I'll be using to make the final version.

Screenshot from 2020-06-07 17-11-46

maxtron95:
@Foggalong Thanks! I was trying to make the lines fit on the pixel grid when I was designing that icon.

Foggalong:
@creepertron95 When they're all diagonal lines like that, making them consistent throughout the design will be more important than having nodes on grid because they'll be blurry anyway.

Foggalong:
Final versions for reference :tada: 

CIRCLE SQUARE

Foggalong (approved):

PR #5321, opened 2020-05-31
Icons for 3DCoat

maxtron95:
https://3dcoat.com/

3dcoat

3dcoat-ci
3dcoat-sq


palob:
What do you think about this simplified version (shadow of the original path for comparison)

simple

maxtron95:
@palob Looks good. It is much better in terms of pixel alignment and uses less nodes.

palob:
OK, I'll push it.

palob (approved):

Foggalong:
Added a v. light blue background circle to round out the icon a little more but very much like the aligning of the drips, lovely touch :+1:

PR #5325, opened 2020-06-01
updated GTK logo

palob:
old
gtk-old
new
gtk-upd

updated the gtkhash icon accordingly
gtkhash
gtkhash-old
gtkhash-upd


Foggalong (changes requested):
I like the GTK update but I think the gtkhash design needs some work, though that's not helped by the new upstreamdesign being nooooot great. Is there a way to tweak the design so that it doesn't look like it has an outline?

palob:
My first idea was to just slap a plain right-angled **#** on it but then I thought the idea of the original one of having the **#**-lines following the ones of the GTK cube was not the worst as it makes the icon look less cluttered. It does work without the meme-style outline but the **#** will look a bit flimsy then.

palob:
I'll try how it looks with the outline turned white.

palob:
Does this look better?
gtkhash2

Foggalong:
Yeah, like that much more!

palob:
Pushed.

Foggalong (approved):

PR #5326, opened 2020-06-01
Icons for Trinity Desktop Environment

maxtron95:
https://www.trinitydesktop.org/

tde-logo

tde-ci
tde-sq


palob (approved):
I find this a strange project but the icons are good to go.

palob:
What's the ``go`` symlink for? Not sure if an app icon named like this is a good idea.

maxtron95:
The go symlink is an icon from the default icon theme in TDE. (https://wiki.trinitydesktop.org/Icon_Sets)

palob:
Do you know which app this is supposed to be for?

maxtron95:
I am thinking it is one of the system icons for TDE.

Foggalong (changes requested):
Yeeeah veto on the icon called `go.png` just being a symlink to TDE, remove that from the JSON data. Aside from it being a bad idea to name an icon that, I've looked it up in [the source](https://github.com/thecodingmachine/crystal-project) and there's no mention of where it originally came from and the icon itself is both nondescript _and_ nothing like the Trinity logo: go.

The symbol on the circle icon also needs moving down 1px I think, looks too top heavy as is.

Foggalong:
> @palob: I find this a strange project

I suppose it's not that dissimilar to MATE in its aim to keep the GNOME 2 desktop paradigm alive when GNOME 3 went in a different direction. That said, MATE seems to have made a conscious effort to make sure it keeps up with modern technology and design while Trinity doesn't just _look_ dated but seems to be using very old packages to make that happen too (Qt3 on which it's version of Qt is based was released nearly 20 years ago, and it's 15 years since Qt4 was released).

maxtron95:
@Foggalong Since the go icon is generic, maybe it can be symlinked onto something like numix.svg? I'm thinking the go icon is used as an about icon like `distributor-logo`.

That icon you found is actually the logo of a linux distro called Linspire. There is a go icon that uses the Trinity logo in the default icon theme of Trinity.

There is one linux distro that uses Trinity by default and that is Q4OS. (Which I made icons for, but haven't made a PR for yet.)

palob:
There are several commonly used ``go-...`` action icons and I've seen KDE/Qt apps picking up icons (as fallback) despite the icon name not being complete so I don't think we should have this included in our app icon themes. I guess the icon is used for some welcome screen, so the launcher probably has ``NoDisplay=true`` set and you won't see the icon ever again after first boot.

Foggalong:
Yeah it's going to be a hard no on including a `go.svg` even as a generic symlink, chance of it opening us up to future bugs isn't worth the risk. Just pushed a commit to delete that symlink.

@creepertron95 for context if you need it at a later date, our go-to icon for icon names which are too generic would be `icon.svg` (just three cogs) not `numix.svg`. This isn't a case of that though; here the icon name is such it could be accidentally picked up as an action icon.

palob (approved):

palob (approved):

Foggalong (approved):

PR #5327, opened 2020-06-01
Icons for 7 Kingdoms: Ancient Adversaries

maxtron95:
https://7kfans.com/

7kaa

7kaa-ci
7kaa-sq


palob (changes requested):
I think the bezel colour should me much brighter, like below or even more so. Right now it looks like a miscoloured shadow.

bezel

palob (changes requested):
Symbol on circle should sit 1px higher, maybe same for Square. Let's do it for now.

You got your icons misnamed with a ``-ci`` prefix at the end. This is why the build fails.

palob (changes requested):
Square icon is still misnamed.

Foggalong (approved):
I'm good with this now! @palob?

palob (approved):

PR #5334, opened 2020-06-09
Icons for Alien Swarm games

maxtron95:
Alien Swarm - 

https://store.steampowered.com/app/630/

alien-swarm

as-ci
as-sq

---

Alien Swarm: Reactive Drop - 

alien-swarm-reactive-drop

asrd-ci
asrd-sq


palob (changes requested):
The original alien-swarm symbol looks a bit more blueish/livelier. Could we get something like this?
The original reactive drop icon looks slightly brighter.

Foggalong (changes requested):
I actually quite like the blue of Swarm but agreed, Reactive drop needs to have a brighter symbol

palob (approved):

Foggalong (approved):

PR #5336, opened 2020-06-09
Icons for Amorous

maxtron95:
https://store.steampowered.com/app/778700/

amorous

amorous-ci
amorous-sq


palob (changes requested):
Both icons have the shadow misplaced vertically.

Both icons should have the symbol moved down by 1px.

PR #5339, opened 2020-06-09
Icons for KawAnime

maxtron95:
https://github.com/Kylart/KawAnime

kawanime

kawanime-ci
kawanime-sq


palob (changes requested):
Symbol should be shortened a bit and the person looks a bit much gothic/zombie-like/past their prime.

Much of this impression is due to the eyes.

maxtron95:
@palob Any advice for the eyes?

palob:
Well, they shouldn't be black(ish), have pupils/irises and more manga-like proportions.

palob:
In general view the pen looks like a weapon, so maybe the pointy end should be on the left.
The left arm looks a bit long because the tablet thing is a bit tall.

palob:
The zipper gap looks like a pointy chin or a beard.

maxtron95:
Modified Icon:

kawanime-new-ci

palob (commented):
Thanks, that helped a lot! Does she look sulky (->mouth) on purpose?
Could you try a baseplate with more contrast to the face?

maxtron95:
@palob Should I try a more dark-grayish baseplate?

palob:
If you find a shade with good contrast to the jacket. Maybe something sky blueish if that doesn't work.

palob:
Eyes are too complicated as is.

PR #5341, opened 2020-06-09
Icons for MiKTeX console

palob:
https://miktex.org/howto/miktex-console

miketx-console
miketx-console-ci
miketx-console-sq


Foggalong (approved):

Foggalong (changes requested):
I think we can maybe make this more detailed with an actual grey spanner, rather than the outline

Issue #5153, opened 2020-03-30
Phase-Out Browser Baseplate


Foggalong:
The browser baseplate we use across a wide range of icons is starting to look significantly more dated than it did however many years ago I chucked it together. This issue is a general note to me (or to anyone else who fancies a go) to try and come up with an alternative which ideally:

1. Can still be used in a wide variety of icons
2. Looks nicer while not just focusing in Europe or America
3. Isn't a full baseplate design, since those are being phased out generally

palob:
I was about to open an issue about this for technical reasons.

Foggalong:
Are the technical reasons how ridiculously complicated it is? :joy:

maxtron95:
Well there is a lot of web browser icons (including firefox) that is using the browser baseplate. I think the firefox icon looks weird as a full baseplate square icon in my opinion.

Foggalong:
@creepertron95 Yeah full baseplate icons are being phased out generally, the browser one in particular though is one which has been niggling me because it's a poor (and complex) design.

maxtron95:
These are the icons that are currently using the browser baseplate:

- aarddict
- abrowser
- drawpile
- eiskaltdcpp
- empire-total-war
- ephemeral
- eric-web
- firefox-icon-unbranded
- gottengeography
- jamulus
- luakit
- midori
- mlclient
- mlterm
- nmap
- ooo-web
- pac
- parley
- pglgui
- qmaptool
- qomp
- qrestclient
- qutebrowser
- smtube
- smuxi-frontend-gnome
- web-browser
- zeroinstall

maxtron95:
@palob Seems we can ditch the globe on the `komikku` icon. The icon has just been redesigned since you last made a PR for it in #5053.

![info febvre Komikku](https://user-images.githubusercontent.com/53346722/81417357-ab707680-9110-11ea-845b-0ff7a2a7ae86.png)

palob:
The lowest hanging fruit-solution would be a (symbol-on-baseplate) globe with fantasy continents on it. If we ditch the fullscreen design, we need to find ways to incorporate the on-globe-baseplate-symbols.

Right now we use the globe design for different kinds of applications: web browsers, map/navigation apps, communication apps, localisation apps, games.

There are other widely-used ambiguous metaphors, for instance a compass but maybe we won't be able to find a new one which fits all.

Question is if the globe metaphor is a good one for all the apps in the first place.

Foggalong:
[...]

PR #5349, opened 2020-06-13
Icons for OSM Scout Server

palob:
https://github.com/rinigus/osmscout-server

Contender for the worst icon in existance
harbour-osmscout-server
harbour-osmscout-server-ci
harbour-osmscout-server-sq


Foggalong (commented):
I'm holding this one until we've come up with a  simplified world map (as part of #5153) to use for icons like this where it makes more sense to have a real globe rather than a fictional one.

Foggalong:
@palob I'm using :label: wontfixto as a proxy for it being "on hold", it's not saying that this won't be merged at some point :slightly_smiling_face:

palob:
The wire globe doesn't seem to work that well here.
osmscout-server

palob:
Might work with lines on a circle or the Epiphany style.

palob:
Since this is about map tiles, why not something with tiles?
I'm not fond of this particular draft though.
osmscout-server-tiles

Didn't come up with a good metaphor yet.

Foggalong:
That's a decent shout yeah! Would increasing the size of each tile work (say a 4x4 grid rather than 5x5)

palob:
That's a good idea but will need more experimenting to get a even distribution of tiles / centered symbol.

palob:
Do you think this is any good, @Foggalong? I'm still not sure.
osmscout-server

palob:
Solid-colour tiles or this one. 
osmscout-server

PR #5350, opened 2020-06-13
Dedicated icons for marble

palob:
I'm thinking of something globe-like. Not quite sure this is it, might be just the colours.

marble-ci
marble-sq

alt:
marble-2


Foggalong:
I'm holding this one until we've come up with a  simplified world map (as part of #5153) to use for icons like this where it makes more sense to have a real globe rather than a fictional one.

Foggalong (commented):
I'm holding this one until we've come up with a  simplified world map (as part of #5153) to use for icons like this where it makes more sense to have a real globe rather than a fictional one.

palob:
How about a simplistic approach as this one?
marble3g
#5114 could be symlinked or slightly altered.
#5349 could be implemented as this wire globe or a simple circle on top of the connector.

palob:
marble-new1
marble-new2

What do you think, @Foggalong ? Any good?

Foggalong:
Even though it's an align-to-grid-nightmare I absolutely love that second one!

palob:
Will try how it looks with the axis upright.

palob:
Ah, it doesn't work with the axis upright due to the very reason of alignment. Strokes need to be 2px then which doesn't look good at all.

palob:
Got another idea I'll give a shot in the next days: latitude/longitude lines on top of a circle instead of a "hollow wire globe".

palob:
Found a way to simplify this a bit more.
marble0

Maybe the polar circles should be slightly bent.

palob:
I prefer lines on a solid circle, looks less busy and more like an actually usable globe.
marble0b2
marble0b3

palob:
No I could go "so far" and do something like in the new [GNOME Web aka Epihany](https://gitlab.gnome.org/GNOME/epiphany/-/blob/a308ad5a61b63443214b65a388d9910573567d1b/data/icons/hicolor/scalable/apps/org.gnome.Epiphany.svg) icon, differently coloured surfaces instead of lines, probably more Numixy.

palob:
Here's the Epiphany style. Maybe too 3D-ish?
marble-epiph4
marble-epiph5

Any preference, @Foggalong ?

palob:
Here are the latest contenders, @Foggalong 
marble-var1
marble-var2
marble-var3

palob:
We could use the first design (without stand and tilt + using original colours) for a redesigned Epiphany icon.

I'll need some more experimenting to see if it can replace the icon using the browser baseplate or incorporating a globe symbol.
Strictly speaking this design would require distinct colouring for Circle and Square to cater for the respective light sources.

Since the symbol is tilted here, simply flipping horizontally will probably do the trick in this case.

Foggalong:
I think the first and second designs look a bit too 3D b/c of the curvature. Out of curiosity, how does the first design look when the lines are straightened out as the third design?

palob:
Finally some progress here,@Foggalong. Globe design, straightened out latitudes.
marble-work7

palob:
Pushing this one. I still think it is borderline 3Dish but it doesn't work with just a solid-colour circle.

marble-work8

PR #5352, opened 2020-06-14
Icons for Alien Blaster

maxtron95:
http://www.schwardtnet.de/alienblaster/

alienblastericon

alienblaster-ci
alienblaster-sq

### DIFF
--- a/data.json
+++ b/data.json
@@ -598,6 +598,11 @@
             ]
         }
     },
+    "alienblaster": {
+        "linux": {
+            "root": "alienblaster"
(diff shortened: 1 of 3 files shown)


palob (approved):
Just if you've got some time to waste: there are a few ``paint-order:stroke fill markers`` instances.

Foggalong (changes requested):
To add to the fill markers, this symbol looks a bit small to me. Maybe having it horizontal would allow some more space to expand it some?

PR #5356, opened 2020-06-16
Icon for Witcher 3

ntcarlson:
| --- | :-- | :-- |
|steam_icon_292030| witcher3_circle | witcher3_square|


palob (changes requested):
Hi, thank you for your contribution, it is a very lovely design. There a few clashes with our design guidelines though:

+ The symbol is a bit big. It is somewhat circular (those can be up to 30px X 30px) but I think it should be no wider than 26px. Unfortunately you'll have to rework the pixel grid alignment after resizing.
+ We try to avoid on-symbol gradients for new additions. Can you find another way to implement nose + maw?
+ The circle icon has the baseplate gradient gone awry. It should be placed liked this.
gradient
+ The black parts of the symbol look too dark and should be lightened up.
+ The baseplate gradient doesn't contrast well with the symbol. It should be either slightly darker, a lot brighter and/or use a different colour hue.
+ There's some SVG cruft in your files. Could you please run search&replace ``;font-variation-settings:normal;stop-color:#000000`` in a text editor?

ntcarlson:
Thanks for the detailed feedback. I think I addressed all your concerns with my latest commits.

palob (changes requested):
Thank your for your changes.
Small issues left:
+ There are greyish duplicates behind the eyes.
+ The baseplate-symbol-contrast still isn't good and the baseplate darkness obfuscates the symbol shadow. Here's a (more or less randomly choosen) colour combination that could work. Feel free to find one you like which fits to the game's theme.
colour

+ Could you please cut off the lower teeth path like this? Otherwise they'll shine through due to hw SVG rendering works.
teeth
+ The symbol on the Circle icon should sit 1px further down.

ntcarlson:
Ok the small issues should be fixed. I also took another stab at the baseplate color, but I don't really have a good eye for that sort of thing so if it still doesn't look quite right, I'll go ahead and use your suggestion

palob (commented):
Thank you for the changes. Now with the brighter baseplate the symbol elements should be brightened up a little. I can push these chnges if you don't mind, might be easier than telling you the colours for each path.

ntcarlson:
Sure go for it

palob:
Alright but I'll wait for @Foggalong's review in order to avoid duplicate effort.

Foggalong (changes requested):
Aye, agree regarding the lighter brown it's a much nicer contrast! Great first design though @ntcarlson!

palob:
Pushed these
witcher-ci
witcher-sq

Not easy to get right. Tried to incorporate the red from the game logo, to no avail.

palob:
Pushed these
witcher-ci
witcher-sq

Not easy to get right. Tried to incorporate the red from the game logo, to no avail.

Foggalong (approved):

maxtron95:
@palob Should we try doing the game icons for Witcher 1 and 2 as well?

palob:
If you fell like doing so, I won't stop you :-)

PR #5361, opened 2020-06-24
Icons for 4digits

maxtron95:
https://github.com/fossilet/4digits

4digits

4digits-ci
4digits-sq


palob (approved):

Foggalong (changes requested):
The design for this is good, but the baseplate colour needs adjusting so that it contrasts better with the symbols.

PR #5364, opened 2020-06-26
Icons for A7Xpg

maxtron95:
http://a7xpg.sourceforge.net/

a7xpg

a7xpg-ci
a7xpg-sq


palob (changes requested):
Nice creative design!
Could you please do something like this to counteract the outline rendering issue?
outlines

Foggalong (dismissed):

Foggalong:
I think the baseplate has to be a bit lighter, as is you can't see the baseplate shadow :(

Foggalong (changes requested):
I think the baseplate has to be a bit lighter, as is you can't see the baseplate shadow :(

PR #5365, opened 2020-06-26
Icons for Android ADB File Manager

maxtron95:
https://github.com/sole/aafm

aafm

aafm-ci
aadb-sq


palob (approved):

Foggalong (changes requested):
I'm not so certain on using a cut-out here rather than a file manager type symbol. Maybe something [like this](https://github.com/numixproject/numix-core/blob/master/icons/circle/48/thunar.svg) but with the Android head?

As an aside for both designs, any icon which uses a folder symbol also needs sticking in the [Numix Folders](https://github.com/numixproject/numix-folders) repo with alternate designs for each of the folder types we offer.

maxtron95:
If that is the case, I can also add in an Android Folder to the base Numix theme as well.

palob:
I'm merging this as is. Upstream is abandoned so this isn't worth any extra effort.

PR #5369, opened 2020-06-26
Icons for Anna's Quest

maxtron95:
https://store.steampowered.com/app/327220/

annas-quest

annas-quest-ci
annas-quest-sq


palob (changes requested):
Facial features should be stacked like this and be a bit darker due to the fine-grainedness
facial

There is one SVG cruft block left in the icons.

palob (approved):

PR #5370, opened 2020-06-26
Icons for Abe's Amazing Adventure

maxtron95:
http://abe.sourceforge.net/

abe

abe-ci
abe-sq


palob (changes requested):
Compared to the original icon the guy looks a bit too obese, especially in the back.
Arm posture is a bit strange, the arm doesn't seem to be connected with the shoulder.

palob (changes requested):
Yeah, that looks more like ole Abe.
I think however the background should be brighter.

PR #5405, opened 2020-06-30
Runtime Steam Icons fix

BlueManCZ:
I recently wrote this python script, to allow users to fix their runtime Steam icons. Some people may find it useful as I do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ We accept contributions, read [this page](https://github.com/numixproject/numix-
 ### Hardcoded Icons
 To deal with hardcoded application icons Numix uses the [hardcode-fixer](https://github.com/Foggalong/hardcode-fixer) script. A list of the applications supported by the script can be found [here](https://github.com/Foggalong/hardcode-fixer/wiki/App-Support).
 
+### Steam Runtime Icons
+To fix icons of running Steam games, you can use [Steam Icons Fixer](https://github.com/BlueManCZ/SIF) script, that will connect all icons available from our icon theme with your installed games.
 
 ## Script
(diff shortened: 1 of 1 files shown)


palob (approved):

Foggalong:
This repository is for the themes' build environment rather than the theme packages themselves so it wouldn't be appropriate to list it here. This looks like a brilliant script though so I'd love it if you would submit a notice similar to the "Hardcoded Icons" section of the READMEs in our [Circle](https://github.com/numixproject/numix-icon-theme-circle) and [Square](https://github.com/numixproject/numix-icon-theme-square) package repos so I can merge it there.

For context, the reason [hardcode-fixer](https://github.com/Foggalong/hardcode-fixer/) is listed in this repo is that in fixing launchers its uses its [own list of icon names](https://github.com/Foggalong/hardcode-fixer/blob/master/tofix.csv) which are distinct from the names used by the original app. As an example, for DupeGuru it changes the icon line from `Icon=/usr/share/dupeguru_se/dgse_logo_128.png` to `Icon=dupeguru`. There isn't a standard icon called `dupeguru` so themes still have to create a new icon with that name. By having hardcode fixer listed here it just signifies that we also explicitly aim to cover the icon names which only come only from that script as well.

On the other hand, your fixer uses the standard icon names `steam_icon_$appId` which all icon themes aim to cover as part of their normal business since that's also the format Steam `.desktop` launchers use too. If a theme supports Steam apps it de-facto also supports your script. The same isn't true for Hardcode Fixer since extra icons are needed on the part of theme developers.

PR #5408, opened 2020-06-30
Icons for Airstrike

maxtron95:
https://icculus.org/airstrike/

airstrike

airstrike-ci
airstrike-sq


palob (changes requested):
The stripes look somewhat odd. How does it look with simple parallel stripes?

PR #5409, opened 2020-06-30
Icons for Airtame

maxtron95:
https://airtame.com/

airtame

airtame-ci
airtame-sq


palob (changes requested):
Symbol is slightly asymmetrical. Could you please rebuild this from a 3px stroke? Warning, this will cause cruft.
stroke

PR #5413, opened 2020-06-30
Icons for AlienFX

maxtron95:
https://github.com/trackmastersteve/alienfx

alienfx

alienfx-ci
alienfx-sq


palob (changes requested):
Could we please have the background a tad brighter?

PR #5414, opened 2020-07-01
Icons for Antihero

maxtron95:
https://store.steampowered.com/app/505640/

antihero

antihero-ci
antihero-sq


palob (changes requested):
This could be a little brighter altogether and has a cruft block left.

PR #5415, opened 2020-07-01
Icons for AppGameKit applications

maxtron95:
https://www.appgamekit.com/

- AppGameKit Classic

appgamekit-classic

agk-ci
agk-sq

- AppGameKit Studio

appgamekit-studio

agk-s-ci
agk-s-sq


palob (changes requested):
At small sizes this can be confused as **ACK**.
Could you please make the G hook more prominent (perhaps taller)?

PR #5416, opened 2020-07-01
Icons for Aquaria

maxtron95:
https://store.steampowered.com/app/24420/

aquaria

aquaria-ci
aquaria-sq


palob (changes requested):
Could we please have a little more contrast for the facial features and maybe the eyes aligned like this?
eyes

The build failure isn't caused by you it appears.

PR #5421, opened 2020-07-01
Icons for Altus

maxtron95:
https://github.com/amanharwara/altus

altus

altus-ci
altus-sq


palob:
How would it look with just the bird on baseplate?
If we decide to keep this "squared circle" it needs to go 1px up and to the right.

maxtron95:
Should I go ahead and ditch the "squared circle"?

palob:
I'm excited to see whethr it works without, the symbol might be too small then.

Foggalong (commented):
I'd be fine to merge this design but agreed, curious what one without the droplet design would look like

PR #5430, opened 2020-07-01
Icons for Another Redis Desktop Manager

maxtron95:
https://github.com/qishibo/AnotherRedisDesktopManager

ardm-ci
ardm-sq


palob (changes requested):
Both symbols should be moved 1px to the right and both files include cruft blocks.

PR #5436, opened 2020-07-01
Icons for APKTool

maxtron95:
https://ibotpeaches.github.io/Apktool/

kali-apktool

apktool-ci
apktool-sq


palob (changes requested):
It has those outlines and the symbol is just the app name written out but it is at least a creative typeface so I'd be OK with that. 
Could you please remove the cruft?

Foggalong (changes requested):
Can we get a non-white baseplate colour as well?

PR #5440, opened 2020-07-02
Icons for Warframe

BlueManCZ:
| --- | --- | --- |
| ![orig][orig] | ![circ][circ] | ![sqr][sqr] |

https://steamdb.info/app/230410/

[orig]: https://user-images.githubusercontent.com/17854950/86353944-6d9f5280-bc68-11ea-8296-5afdbfcea930.png "Original"
[circ]: https://user-images.githubusercontent.com/17854950/86354099-b48d4800-bc68-11ea-9a58-45dcb119ca29.png "Circle"
[sqr]: https://user-images.githubusercontent.com/17854950/86354189-dab2e800-bc68-11ea-8cba-45b0bcf1d834.png "Square"


palob (changes requested):
I'm afraid the symbols are too wide. I see it's going to be difficult to squeeze that design even more, how about only implementing the inner part?
There's something wrong with the baseplate bezel and the baseplate have no vertical colour gradients. It's probably best to start with fresh baseplates which can be found in ``48.svg`` in the ``templates`` directory.

Instead of full-blast white, ``fafafaff`` (98% L) ist the brightest we use in our icons.

BlueManCZ:
Well, I'm afraid if I remove side shapes and implement only the inner part, it wouldn't feel right. Couldn't we make an exception as in [preferences-system-privacy.svg](https://github.com/numixproject/numix-core/blob/master/icons/circle/48/preferences-system-privacy.svg) or [utilities-system-monitor.svg](https://github.com/numixproject/numix-core/blob/master/icons/circle/48/utilities-system-monitor.svg)? Because when you look at other implementations, this logo is really expected to fill the baseplate from right to left (if you know what I mean).

palob (changes requested):
Could you please make small adjustments for better alignment like this? Doesn't need to be exactly like this, just to get the idea.
align

It seems possible to reduce the width of the inner part by 2px by moving the outer nodes 1px inwards and upwards.

Please merge all shadow paths into one single path.

There's lots of instances of ``;paint-order:stroke fill markers`` in your files which is basically cruft. Could you please remove this (text editor + search&replace should do it)

palob (changes requested):
Thank you, you even got it down to 30px :+1: 

Regarding alignment, there's a few lines very close to pixel borders which could be made straight or parts which only very slightly exceed a border. Dou you understand what I mean?

BlueManCZ:
@palob Well, I understand what you mean, but I'm not sure which lines you mean.

BlueManCZ:
Something like this?

**Edit**
Ah, now I see you already did it on your screenshot above.

Snímek z 2020-07-02 21-09-49

Foggalong (changes requested):
I'm good with being a little bit lenient with the width since this is a Steam game, but can we make the colouring a bit more unique? While the default logo version is the white or blue solid version there seems to be other official uses from expansions and press releases which have a lot more colour to them.
1. wf1
2. wf1
3. wf1
4. wf1

Incorporating any of those into this would make it a bit more lively, even if it's just having slightly darker outer shapes and slightly lighter inner shapes.

BlueManCZ:
@Foggalong  What do you think?

export2

palob:
I'm all for more colours. Now with only the center part white the symbol looks almost small.

Foggalong:
Yeah when I downloaded the updated version it looked small compared to the all white one, but sticking it alongside some other icons it doesn't look too bad. Think it's just an illusion of the colour change?

BlueManCZ:
I reduced the width of the symbol to 30px in https://github.com/numixproject/numix-core/pull/5440/commits/a7b188407e07a4ff80507d658c620e457976739f because @palob pointed out it was too wide. But since then I didn't change it. So as you say, it must be an optical illusion. I can extend the width again if you both agree.

Foggalong:
@palob Happy to follow your lead on this one!

palob:
I'd say 32 px for the width (intermediate between now and the first draft) + the new colours.

BlueManCZ:
|A|B|C|D|
|---|---|---|---|
|A|B tmp|C|D

I made some drafts. C, D are probably a little too much, right? Should I stick with default shapes in A?



But I have a feeling that the more colours we add, the worse it will look.

BlueManCZ:
Or maybe rotate basic colours like this?
|E|F|G|
|---|---|---|
|E|F|G|

palob:
For me A and E are the contenders, I vote vor E.

BlueManCZ:
war

palob (changes requested):
Symbol shadow in the Square icon needs to be shifted 1px to the left.

palob (approved):

Foggalong (approved):

PR #5442, opened 2020-07-02
Icons for ARAnyM

maxtron95:
https://aranym.github.io/

aranym

aranym-ci
aranym-sq


palob (changes requested):
There's a few areas where alignment and rounding isn't optimal which contributes to the overall wobbly look.

maxtron95:
@palob Should I try to make the icon an Atari Computer like [this](https://upload.wikimedia.org/wikipedia/commons/thumb/3/39/Atari_1040STf.jpg/800px-Atari_1040STf.jpg) instead, since this an Atari virtual machine software?

palob:
This might work, maybe incorporating the colours of the original icon or something.

maxtron95:
@palob What do you think of this new design?

aranym-ci

palob:
Looks ace but maybe a bit big and detailed.

Foggalong (changes requested):
Agreed new design is better but too big; possibly will need to be 1-2px higher after being resized also

PR #5445, opened 2020-07-02
Icons for Dungeon of the Endless

mrmeszaros:
| --- | :-- | :-- |
| dungeon-of-the-endless original | dungeon-of-the-endless circle 192 | dungeon-of-the-endless square 192 |


palob (commented):
Does this look any good without perspective, i.e. vertically symmetrical?

Symbol should be slightly smaller, 28x28 and the baseplate brighter so that symbol shadows and baseplate bezels are esier to recognise.

mrmeszaros:
In all the base is brighter, the symbol is smaller and has less perspective.

| A | B | C | D |
|---|---|---|---|
| dungeon-of-the-endless circle 192 | dungeon-of-the-endless circle 192 | dungeon-of-the-endless circle | dungeon-of-the-endless circle 192 |

- B has even brighter base
- C is without _no perspective_ - personally it is a bit bleak for me.  
- D is vertically mirrored

I think D or B.

palob:
D it is :+1:  because judging by the shadows for Numix Circle the light source is at the top left, for Square at the center top.

Foggalong (commented):
What's the blue diamond supposed to be? I'm not sure from the pixelated original icon what it's supposed to be.

palob (approved):

mrmeszaros:
The yellow crystal is some sort of power source, however, I have not seen the blue diamond in the game.
To me it looks like the crystal's peak broke off, and it is the insides.

Foggalong (approved):
Ahhh I see it now. The perspective for that is slightly off in the blown up version which is what threw me, but it works at any reasonable size so all good :+1:

PR #5446, opened 2020-07-02
Icons for Augur

maxtron95:
https://github.com/AugurProject/augur-app

augur

augur-ci
augur-sq


palob (commented):
Is it possible to lower the circle such that the horizontal lines meet the corners of the pyramid? Maybe the corners need to go up.

Foggalong (changes requested):
All the lines here should be the same thickness, it seems very off compared to the original having the diagonal and straight lines be different. 

Also agreed with @palob, would be curious how it looks with the horizontal lines emerging from the corners.

PR #5447, opened 2020-07-02
Icons for Baiji Manga Viewer

maxtron95:
https://github.com/jmarioste/manga-viewer

baiji-manga-viewer

baiji-ci
baiji-sq


palob (changes requested):
There's a very small broken spot
broken
and the main body should retract underneath the white path to prevent outlines.

palob (approved):

PR #5450, opened 2020-07-02
Icons for BloomRPC

maxtron95:
https://github.com/uw-labs/bloomrpc

bloomrpc

bloomrpc-ci
bloomrpc-sq


palob (changes requested):
The petal could be better aligned, like this
align
Either using 4 or 2 nodes for this kind of petal.

Could we please get a greener shade for the right petal. Right now the left green one is a bit lost and the overall greenness of the original icon feels underrepresented.

PR #5452, opened 2020-07-03
Icons for Sid Meier's Civilization VI

BlueManCZ:
| Original icon | Circle icon | Square icon |
| --- | --- | --- |
| ![orig][orig] | ![circ][circ] | ![sqr][sqr] |

https://steamdb.info/app/289070/

[orig]: https://user-images.githubusercontent.com/17854950/86492457-7c702d00-bd6e-11ea-89ac-7e94fcf6f885.png "Original"
[circ]: https://user-images.githubusercontent.com/17854950/86492473-85f99500-bd6e-11ea-8f1b-acece9f76240.png "Circle"
[sqr]: https://user-images.githubusercontent.com/17854950/86492598-ef79a380-bd6e-11ea-89c9-a5bd3293f5c8.png "Square"

PR #5453, opened 2020-07-04
Fix for Sid Meier's Civilization V

BlueManCZ:


Foggalong (changes requested):
TIL we had two versions of the icon in the repo, good catch! This is probably as good an opportunity as any to follow on from the discussion of #5452 and see whether a slightly lighter 'V' inset would look like as per the original icon.

BlueManCZ:
I'll look into it. :wink:

BlueManCZ:
civ5

palob (approved):

PR #5454, opened 2020-07-04
Icons for Cacher

maxtron95:
https://www.cacher.io/

cacher

cacher-ci
cacher-sq


Foggalong:
I've watched that gif on their website loop like five times and was still no clearer as to what their app actually did :joy:

maxtron95:
Oh, I accidentally pushed another icon here sorry.

palob (changes requested):
I think ``<``, ``/`` and ``>`` should be either of the same height or they should be further apart.

palob (approved):

PR #5456, opened 2020-07-04
Icons for Conky Logomark Violet

maxtron95:
https://github.com/brndnmtthws/conky

conky-logomark-violet-2

conky-sq
conky-ci


palob (changes requested):
Only relevant for Circle: The inner blue rectangle doesn't cast a shadow on the baseplate.
The white baseplate looks a bit harsh here. Could we please get a slight bright blueish colouration?

maxtron95:
I modified the baseplate, I was going with a white baseplate to match the other conky icons.

palob (approved):

Foggalong (changes requested):
Looking at the repo this application _is_ conky, so this PR should be set as a proposed update to the existing conky icon with `conky-logomark-violet` as a symlink.

palob:
I'll adjust this one as the new ``conky`` one while the existing one will be the new ``conky-manager`` one.

PR #5462, opened 2020-07-04
Icons for Cryptr

maxtron95:
https://github.com/adobe/cryptr

cryptr

cryptr-ci
cryptr-sq


palob (changes requested):
Could we have the inner plate a bit brighter like in the original?

palob (approved):

PR #5464, opened 2020-07-04
Icons for Aragami

maxtron95:
https://store.steampowered.com/app/280160/

aragami-icon
aragami

aragami-ci
aragami-sq


palob (approved):

Foggalong (changes requested):
Could we maybe get some darker red folds in the hood a la the original?

PR #5467, opened 2020-07-05
Icons for CutePeaks

maxtron95:
https://github.com/labsquare/CutePeaks

cutepeaks

cutepeaks-ci
cutepeaks-sq


palob (changes requested):
Symbol could be a bit bigger. In total this one can be sized up to 28x22.
The 2 left wave could cast a shadow onto the right ones.

PR #5470, opened 2020-07-05
Icons for Cyan

maxtron95:
https://github.com/rodlie/cyan

cyan

cyan-ci
cyan-sq


palob (changes requested):
Symbols should be 1px higher, the baseplate should be a bit darker.
Could you please try to improve alignment of the **C** symbol?

PR #5474, opened 2020-07-06
Icons for dManager

maxtron95:
https://dmanager.stevertus.com/

dmanager

dmanager-ci
dmanager-sq


palob (approved):

Foggalong (changes requested):
Not a fan of having this as a 3D symbol. How does it look with just the d as a flat symbol in the hexagon?

Foggalong:
Also an aside, but that is a terrible website; it doesn't explain anywhere on the main page that it's for Minecraft!

PR #5475, opened 2020-07-06
Icons for DockStation

maxtron95:
https://dockstation.io/

dockstation

dockstation-ci
dockstation-sq


palob:
It's a good design but I'm not sure if we are going to include more full-baseplate designs, @Foggalong ?

Foggalong (changes requested):
Yeah, it's a no on new full baseplate designs. I believe the original is supposed to resemble a porthole so I think that's what we could do here; this circle design as the 26px symbol with a 2px white circle "frame" around it (a la the original, making a 30px circular symbol in total) on a grey baseplate.

Issue #5379, opened 2020-06-28
Legacy Android Icons


Foggalong:
Issue #5175 has brought to light that there are a fair few [Square](https://github.com/numixproject/com.numix.icons_square/tree/master/Icons) and [Circle](https://github.com/numixproject/com.numix.icons_circle/tree/master/Icons) Android icons from the legacy themes which were never transferred over to the Core system. It might be worth someone digging through and seeing what from there is worth salvaging.

mrmeszaros:
@Foggalong those links to the square and circle repositories do not work anymore.

palob:
As I don't happen to find the old repos locally this might be unsalvageable

PR #5479, opened 2020-07-07
Icons for uTorrent

maxtron95:
For #5379 

https://www.utorrent.com/

utorrent-ci
utorrent-sq


palob:
db entry needs a comma before ``utorrent``.

palob:
Not sure if this can be considered a full-baseplate design but I think the symbol should be a bit smaller and the upper nodes should be corners, I mean it probably should be a sharp **μ**.

palob:
If this turns out to be too full-baseplatish the most simplistic solution would be a recolouring of one of our ``μ``-icons.

maxtron95:
Note: This icon is from the Numix Android Theme

palob:
Ah, I see, the Circle icon has the baseplate shining through.

Foggalong (changes requested):
Yeeeeeah this might be from Circle's Android version but it definitely wouldn't make it into the theme today as is. Maybe some adjustment to make it a centre symbol [like this](https://idroot.us/wp-content/uploads/2015/03/uTorrent-logo.jpg)?

Foggalong:
@creepertron95 When making a new PR and updating the JSON file I'd encourage you to copy an entry from another app rather than writing one out from scratch. That makes it much less likely schema errors will creep into your submissions.

maxtron95:
On top of that, should I make the icon less full baseplatey and more of a symbol? and should I modify the bittorrent icon as well? (Both uTorrent and BitTorrent are related)

Foggalong:
@creepertron95 Aye modify both :+1:

Foggalong (commented):
As an alternative to [that](https://idroot.us/wp-content/uploads/2015/03/uTorrent-logo.jpg), using micropad as a straight an simple icon might work?

palob:
Yeah, micropad or this one #5604

PR #5480, opened 2020-07-07
Icons for tsMuxerGUI

maxtron95:
https://github.com/justdan96/tsMuxer

tsMuxerGui

tsmuxergui-sq
tsmuxergui-ci


palob (changes requested):
Could we please have the cog shadow onto the inner circle or a shadow for the arrow?
The arrow looks more centered when moved 1px to the right and shortened to a width of 7px.

palob (approved):

PR #5481, opened 2020-07-07
Icons for Abuse (Game)

maxtron95:
https://github.com/antrad/Abuse_1996
https://en.wikipedia.org/wiki/Abuse_(video_game)

abuse

abuse-ci
abuse-sq


palob (changes requested):
Circle icon can have the symbol 1px higher.

palob (approved):

Foggalong (changes requested):
Can we get this without the light grey inside-inset? It removes the 3D effect but I think looks better generally too

PR #5486, opened 2020-07-08
Icons for Electorrent

maxtron95:
https://github.com/tympanix/Electorrent

electorrent-ci
electorrent-sq


palob (commented):
Could you please check how it looks with the magnet and the lines 1px further apart?

Foggalong (changes requested):
Agreed, they're too close to the magnet as is. Either they need moving up 1px or the magnet down 1px (whichever keeps the symbol looking centred).

I think we can add some colour to this icon too, rather than keeping it all green. Just grey for the magnet and red/blue for the poles should do.

PR #5487, opened 2020-07-08
Icons for Axe and Axe Electrum

maxtron95:
https://github.com/AXErunners/axe
https://github.com/AXErunners/electrum-axe

- Axe

axe-ci
axe-sq

- Axe Electrum

axe-electrum-ci
axe-electrum-sq


palob (changes requested):
The db entry for ``axe-electrum`` needs a ``root`` label.

Foggalong (changes requested):
Judging by the original shadows the "arrow part" is raised above the main x, so should cast a symbol shadow. The colour choice for the axe-electrum arrow also makes it look like 3D element; needs tweaking to avoid that.

PR #5488, opened 2020-07-10
Icons for Alan Wake

BlueManCZ:
| --- | --- | --- |
| ![orig][orig] | ![circ][circ] | ![sqr][sqr] |

https://steamdb.info/app/108710/screenshots/

[orig]: https://user-images.githubusercontent.com/17854950/87153145-458e9f80-c2b7-11ea-8f03-bd67e58a12a2.jpg "Original"
[circ]: https://user-images.githubusercontent.com/17854950/87153148-47f0f980-c2b7-11ea-8b27-387186b751d1.png "Circle"
[sqr]: https://user-images.githubusercontent.com/17854950/87153460-cea5d680-c2b7-11ea-94dc-dd269bf840fe.png "Square"


palob (commented):
This definitely would need an exception for the gradient in the symbol. On top of that though the design seems a bit too detailed/fine-grained.

palob:
Could you please try just Alan's (I suppose that is his name) silhouette without the A shape?

BlueManCZ:
Well, yes, silhouette without the A shape is possible, but the player in the game spends most of the time in the dark with a flashlight, so the light background doesn't really fit here.

A

And I can't make the silhouette light, because then it wouldn't be possible to make a flashlight.

palob:
Maybe dark blueish baseplate, greyish (lighter or darker than baseplate) silhouette + flashlight?

BlueManCZ:
I don't know. I think it's getting worse.

A A


And to be honest, I quite like the A-shaped idea, and I personally think we should stick with it. Moreover, the player sees this shape each time the game is started.

BlueManCZ:
In these drafts with white silhouette it lost that "standing in the fog" feeling and it doesn't look very realistic, because when you shine a flashlight against someone, you look dark.

BlueManCZ:
A

palob:
Is there a way to streamline/iconify the silhouette a bit? 
What probably is supposed to be a gun looks like a pry bar, a pipe wrench or a calliper gauge even in the original icon.

Foggalong (changes requested):
I think I agree that it would look better on the "A" backdrop, though maybe with Alan slightly bigger within that than originally committed so that it's a bit more recognisable at smaller sizes.

I'd also argue that the gradients aren't super necessary. The one in "A" background could be replaced by hard cut-offs from one shade to another, a la the rolling hills in the background of the original. The one on Alan could be removed and feet added at the bottom, though even if that looks weird the gradient should be much more localised to only cover Alan's legs rather than his whole body.

BlueManCZ:
I slightly enlarged Alan and reworked gun.
There is a comparison between gradient and cut-offs.

b a

In my opinion, the gradient background looks much better. It better reflects the foggy atmosphere of the game.

Foggalong (changes requested):
> In my opinion, the gradient background looks much better. It better reflects the foggy atmosphere

I'd respectfully disagree, I think you did a good job with the cutoffs and it works really well! Makes for a much more interesting icon design wise and more in line with what we do elsewhere in Numix. Would be very happy to merge that design if you push it :heart_eyes:

palob (changes requested):
Could you please transform the torchlight paths to an outer diameter of 3px and place it at X24,Y26?

The horizontal part of the torchlight arm is ever so slightly misaligned in Y-direction.

palob (approved):
Thank you.

PR #5491, opened 2020-07-14
Icons for application-community

maxtron95:
application-community-ci
application-community-sq


Foggalong:
Is there an upstream example of this icon? The heart isn't the metaphor I would have expected. Also curious, what kind of icons call into this?

maxtron95:
This is actually a category icon. The heart metaphor is usually used for that type of icon. I thought it would be fine since there was a bunch of other category icons that are already present in the numix-core.
point

I was actually trying to for a bit, to bring this icon theme on par (catch up) with Papirus, as I realized it was missing several icons from that theme, but was careful not to add icons that was not supposed to be in the core theme.

Should I use a symlink instead?

Foggalong:
Aye, I got that it was a category but it's not a category I'm familiar with. I'm fine with the inclusion of such an icon but just wondered if you had any examples of its upstream usage?

Also thank you! Papirus have done a good job of expanding their theme so its a good one to be on par with.

maxtron95:
I'm not sure of any upstream examples but the icon is included in some default icon themes. (Elementary Xfce)

Foggalong:
@creepertron95 I've tracked this down and it's not actually an `applications-x` type app category icon but an `application-x` type menu category icon which means it's [covered in base](https://github.com/numixproject/numix-icon-theme/blob/master/Numix/48/categories/application-community.svg) rather than as an app icon in core. It's a subtle difference so easily missed - I didn't work it out until I noticed it was usually cropping up in the `categories/` sub directory primarily rather than `apps/`.

PR #5492, opened 2020-07-15
Icons for Half-Life: Alyx

maxtron95:
https://store.steampowered.com/app/546560/

hl-alyx

hl-alyx-ci
hl-alyx-sq


Foggalong (changes requested):
I'm rejecting this not because I don't think your design is right (I do) but it's made me realise that all the Half-Life icons aren't properly aligned to grid. Will submit a commit to fix that to your branch and then merge this in :+1:

PR #5496, opened 2020-07-17
Icons for Left 4 Dead

maxtron95:
https://store.steampowered.com/app/500/

l4d-ci
l4d-sq

(Note: I recolored the Left 4 Dead 2 icon baseplate to its original colors as well)

### DIFF
--- a/data.json
+++ b/data.json
@@ -11108,6 +11108,11 @@
             ]
         }
     },
+    "left-4-dead": {
+        "linux": {
+            "root": "steam_icon_500"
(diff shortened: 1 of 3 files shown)


Foggalong (changes requested):
Saturation needs taking down on the green baseplate, and lightness possibly adjusting accordingly so that it's still sufficiently dark.

Still undecided on the yelliow, but will rereview with fresh eyes once the green is changed :+1:

PR #5497, opened 2020-07-17
Icons for Killing Floor

maxtron95:
https://store.steampowered.com/app/1250/

killing-floor

killing-floor-ci
killing-floor-sq


palob (approved):

Foggalong:
I could have sworn I'd made an icon for this at some point but I can't find it _anywhere_. Maybe I made it for another theme? Either way, this is great!

palob:
We've got this in another colouring, don't remebmber the name.

Foggalong:
It was the name rather than the design that sounded familiar. I think we had one like this for Plague Inc maybe?

maxtron95:
Yes, I used the Plague Inc icon as a template because Killing Floor uses a bio-hazard icon as well.

BlueManCZ:
I would move symbol in circle icon 1px upside. What do you think?

| Original | Moved |
| - | - |
|  |  |

palob:
From this view it looks a bit more centered indeed.

PR #5503, opened 2020-07-18
Icons for 2050 (Android)

maxtron95:
https://play.google.com/store/apps/details?id=org.mattvchandler.a2050

2050-ci
2050-sq


palob (approved):

Foggalong (changes requested):
I'm not sure on the contrast between that yellow circle and the limegreen of the baseplate. Is there anything else which would work nicer?

maxtron95:
@palob Should I try a white baseplate instead?

palob:
Maybe black is another option, misrepresenting how the app looks like though.

PR #5506, opened 2020-07-18
Icons for Mirror's Edge

maxtron95:
https://store.steampowered.com/app/17410/

mirrors-edge-ci
mirrors-edge-sq


palob (commented):
How would this look with an outline of constant thickness?

Foggalong (changes requested):
Agreed on constant border. I think you can go _much_ lighter on the baseplate blue too to reflect the colours of the game

PR #5513, opened 2020-07-20
Icons for To the Moon

maxtron95:
https://store.steampowered.com/app/206440/

to-the-moon-ci
to-the-moon-sq


palob (changes requested):
I'm going to approve it since it isn't a full-blown full-baseplate icon but could we please get a more moonish tint for the moon? Like in the original icon?

palob:
Now the contrast between the two moon parts is a bit harsh, could you please level the colours a bit?

PR #5517, opened 2020-07-21
Icons for Space Engineers 

maxtron95:
https://store.steampowered.com/app/244850/

space-engineers-ci
space-engineers-sq


palob (changes requested):
Wondering if we could get something more spectacular here.
Anyway, the guy should be 1 or 2px wider such that the gap between the legs can be 2px and aligned.

maxtron95:
@palob Maybe we could look at the game's screenshots for some ideas on something more spectacular?

(https://steamdb.info/app/244850/screenshots)

BlueManCZ:
I don't like the green color. I would go with something from their designs. What do you think?

a 
b

maxtron95:
@BlueManCZ That looks better. I think I might pick the second one with the blue symbol.

maxtron95:
What do you think of this design? (I had to make a 2px gap between legs) (I already pushed the design as well)

se-2

BlueManCZ:
It's an astronaut. 2px gap between legs doesn't look right here. I think 1px aligned should be fine, what do you think @palob?

Snímek z 2020-07-21 18-44-37

BlueManCZ:
circle15

BlueManCZ:
http://colidocs.webly3d.net:5000/images/space-engineers-edit.svg

palob:
Yeah, 1px + alignment is alright. I wonder if this would work with an actual spacesuit instead of just the silhouette.

BlueManCZ:
> I wonder if this would work with an actual spacesuit instead of just the silhouette.

What do you mean?

palob:
A detailed spacesuit design instead of a solid colour.

PR #5521, opened 2020-07-21
Icons for Vocaluxe

maxtron95:
https://github.com/Vocaluxe/Vocaluxe

vocaluxe

vocaluxe-ci
vocaluxe-sq


palob (approved):

Foggalong (changes requested):
I'm not so sure on the circle "mic head" in this, it looks a bit odd with the horizontal line. Could we try it either slanted as in the original or without the line and using the grey circles of the background to make a "more realistic" head.

Foggalong:
@palob This is noice :heart:

PR #5522, opened 2020-07-24
Icons for Tilesetter

maxtron95:
https://led.itch.io/tilesetter

tilesetter-ci
tilesetter-sq


palob (changes requested):
Looks a bit dark and symbol is a bit big. Could you please resize it to 26x22?

PR #5524, opened 2020-07-24
Icons for Classic Racers

maxtron95:
https://store.steampowered.com/app/1037480/

classic-racers-ci
classic-racers-sq


palob (changes requested):
Compared to the original icon this looks a bit tall and steep. Could you try moving the upper nodes down by 2px?

PR #5525, opened 2020-07-24
Icons for applications-education 

maxtron95:
lyricsx-ci
lyricsx-sq


palob (changes requested):
Could you please make the **B** 1px narrower and recentre?
abc

Maybe the C could be a bit narrower as well.

Foggalong (commented):
Don't we already have a chalkboard icon like this?

palob:
Yes, for one of the KDE edu apps I think.

palob (approved):

PR #5526, opened 2020-07-24
Way more symlinks

### LINKED ISSUE #2758: Icon for Accessories category
Hi.
There is no Numix Circle icon for the Accessories category of Mint's Software manager.
Its named `applications-utilities`
Original icon:
screenshot from 2015-12-31 19 00 45
or old gnome theme version:
screenshot from 2015-12-31 19 02 28

maxtron95:
- Fixes #2758


palob:
``assistant5`` needs to go after the ``assistant-...`` entries.

data.json
```diff
@@ -5934,7 +5951,9 @@
                 "2EF4_wordpad.0",
                 "97C1_wordpad.0",
                 "applications-office",
-                "wordpad"
+                "wordpad",
+                "wordview",
+                "wordview.xpm"
```
Foggalong: That's one to be fixed by a hardcode fixer, not by us adding a symlink on our end.

data.json
```diff
@@ -7496,6 +7517,7 @@
             "symlinks": [
                 "accessories-character-map",
                 "applications-fonts",
+                "applixware",
```
Foggalong: I'm not sure this is in the right place? It doesn't seem to be to do with font or character mapping from what I can find

data.json
```diff
@@ -11289,6 +11313,7 @@
             "root": "jockey",
             "symlinks": [
                 "applications-development",
+                "artsbuilder",
```
Foggalong: From what I've found this is [Trinity's graphic design tool](https://centos.pkgs.org/7/trinity-r14-x86_64/trinity-artsbuilder-14.0.8-1.el7.x86_64.rpm.html) - why is it getting a spanner?

data.json
```diff
@@ -14289,7 +14318,10 @@
     },
     "natural-selection-2": {
         "linux": {
-            "root": "steam_icon_4920"
+            "root": "steam_icon_4920",
+            "symlinks": [
+                "spaghetti-armageddon"
```
Foggalong: I see why the symlink, but as we've learned with #5585 this isn't great practice for non-generic icons just because they look similar. Possibly make a copy of that but with colours from the game? Might be a good opportunity to neaten up that icon too.

data.json
```diff
@@ -23077,7 +23122,11 @@
     },
     "web-workflowy": {
         "linux": {
-            "root": "chrome-koegeopamaoljbmhnfjbclbocehhgmkm-Default"
+            "root": "web-workflowy",
```
Foggalong: Root doesn't need to be `web-workflowy`, it can just be `workflowy`

data.json
```diff
@@ -23914,7 +23965,10 @@
     },
     "yandex-browser-beta": {
         "linux": {
-            "root": "yandex-browser-beta"
+            "root": "yandex-browser-beta",
+            "symlinks": [
+                "yandex-browser"
+            ]
```
Foggalong: To make our future lives easier, lets switch this round now so the icon and root are both `yandex-browser` and the symlink is `*-beta`

Foggalong (changes requested):

data.json
palob: By the way, this isn't just a nod to our workflow, adding a double file extension just doesn't work, the icon still won't be shown.

data.json
Foggalong: Given [appixware](https://www.linux.com/news/last-look-applixware/) was reaching EOL in 2006 this is probably moot, but I'm gonna split it off anyways

Foggalong (approved):

PR #5527, opened 2020-07-25
Icons for applications-featured

maxtron95:
applications-featured-ci
applications-featured-sq


palob (commented):
Is there a way to have the horizontal prongs aligned with straight lines between the nodes?

Foggalong (changes requested):
I'd like to see this either use a straight edge design or (and this would be my personal preference, but open to either) using the standard star symbol from [the base theme](https://github.com/numixproject/numix-icon-theme/blob/master/Numix/48/status/starred.svg)

palob (approved):

PR #5528, opened 2020-07-25
Icons for Haskell (applications-haskell)

maxtron95:
https://www.haskell.org/

haskell-ci
haskell-sq


palob:
A general question: where are those applications- icons actually used again? App Stores?

palob (changes requested):
Could we please get the symbol a bit smaller (width reduced to 26px)?

Foggalong (changes requested):
Yeah, I'd like to see where this is actually used? This is a weirdly specific `applications-x` icon and I wouldn't want to accidentally fuck with a poorly named mimetype or something.

PR #5532, opened 2020-07-26
Icons for Atlantik 

maxtron95:
https://www.trinitydesktop.org/applications.php

It is a part of the Trinity Desktop Environment.

- Atlantik

atlantik

atlantik-ci
atlantik-sq

- Atlantik Designer

atlantikdesigner

atlantikdesigner-ci
atlantikdesigner-sq


palob:
Should we add a symlink for gtkatlantic?
https://gtkatlantic.gradator.net/
https://cgit.tuxfamily.org/gtkatlantic/gtkatlantic.git/tree/gtkatlantic.desktop

palob (commented):
It's a bit fine-grained but I don't know exactly what to do about this.

How about filling the street rectangles entirely instead of only the upper part? Looks too far off?

Maybe removing the ouuter and inner outlines could help, leaving only the ones between the "property panels".

Foggalong (approved):

PR #5538, opened 2020-07-27
Icons for esys-escript

maxtron95:
https://flathub.org/apps/details/au.edu.uq.esys.escript
https://github.com/aellery/esys-escript

au edu uq esys escript

escript-ci
escript-sq


palob (changes requested):
I think this would work much better with a triangle grid as symbol. Right now it is not obvious that this is about triangulation, some shapes aren't triangles.

palob:
Or maybe a symbol without outlines but (less) differently coloured triangles.

maxtron95:
@palob Do you know why the check failed?

palob:
Couldn't make it out so far.

maxtron95:
@palob Can you show me what you mean by triangle grid how you envisioned it?

palob:
Well, the lines between the triangles which make out the grid.
A rough estimate would be our ``soundnode`` icon (please disregard the two-tone baseplate:) ).

palob:
@creepertron95 @Foggalong Unfortunately I don't remember how I envisaged that "triangle grid".

I'm not sure the following ideas are recognisable metaphors. Any ideas?
esys1
esys2

Foggalong:
@palob maybe you meant the original symbol as pushed but with more differing colours to make it obvious each shape was a triangle?

palob:
Ah thanks, will try that. Maybe needs reducing the number of triangles.

PR #5541, opened 2020-07-27
Fix icons for Killing Floor

BlueManCZ:
https://github.com/numixproject/numix-core/pull/5497#issuecomment-663240664


palob (approved):

maxtron95:
@palob Should we do the same for the Plague Inc icon?

palob:
Yes, it's a slight difference but I think it looks more centered.

PR #5545, opened 2020-07-28
Add icons for Carrion

mrmeszaros:
| --- | :-- | :-- |
| carrion original | carrion circle 192 | carrion square 192 |

Webpage: https://carrion.game/

A game about being a monster.

Initial version, shadows are missing, @palob could you help with ideas?


palob:
I think the teeth could very well cast a shadow onto the gullet paths. Contrast between the gullet paths could be a tad more intense in order to increase the abyss effect.
The outer teeth could get some "gum bulges" around them like in the original icon.

Will a standard symbol shadow for the red circle make it look like a protruding bulge as seen in the original icon?

palob:
Maybe there's a good looking way to position some of those "plague spots" around the red circle. Symbol will have to be shrinked in order to do that.

Foggalong (changes requested):
Agreed on the teeth, would create a nice inwards effect! I'd be wary about the plague spots making the icon look too busy but open to seeing what it looks like

palob:
Looking at the website background maybe it's easier to do a circular arrangement of "densely packed teeth" on the same level?
Probably easier than getting the 3D effect right.

Foggalong:
That's a good shout!

mrmeszaros:
@palob Good idea!
I will look into it during the holidays!

mrmeszaros:
Hey, I finally had the time have a go at it.
This is what I came up with.
carrion circle 384
I have a few ideas how to create alternatives (like swapping small / big teeth), might create them soon.

palob:
That's a good direction. Maybe a bit simpler shaped teeths and a lower number.

mrmeszaros:
I'll remove the small teeth, that's too much, I agree.
Could You elaborate on the simpler shape?

I thought of tucking the outer teeth under the "gum", to make it less busy / simplify the shapes - like this:

carrion circle 192

Or tuck the inner teeth too?

carrion circle 192

palob:
:+1:  This is way more recognisable as a gullet.

I'm slightly leaning towards variant 1. It is a bit more busy but looks more like a busy whereas variant 2 has a machine/steampunk touch to it. Is there a way to get a shadow for the gum or gum+outer row of teeth?
How does this look with brighter colours?

mrmeszaros:
The shadows help I think (I liked version 1 better with the shadows).
Also, thickening the outer gum helped I think.

carrion circle 384 carrion circle 384 b

Not sure about the colors - the baseplate and the outer gum looks pretty bright to me. Here is a brighter purple / grey / inner teeth for in inside. That might also accentuate the drop shadow too.

palob:
The right one looks ace. Maybe slightly less saturation for the purple.

palob (commented):
Spectacular design, one small nitpick: can you please move the nodes I selected in the screenshot by 0.4 px inwards for better pixel grid alignment

nodes

or by introducing intermediary nodes like this?

nodes2

Is there a way to squeeze some more brightness out of the icon. I think the darkest shade used in existing icons is 18L.

mrmeszaros:
Good catch with the grid alignment! Done.

About the colors - I picked them from the original icon:
| Original | Tweaked | Role |
|---|---|---|
| `#79132e` | as-is | Base |
| `#f10b37` | `#c20a2d` (darker) | Outer Gum |
| `#4d1330` | `#58223e` (desat. + lighter) | Inner Gum |
| `#000000` | `#262626` (lighter) | Throat |

I feel they work pretty well.
However, You are right, the overall brightness is low.
Here two brighter versions:
| Current | B | C |
| - | - | - |
| carrion circle 192 a | carrion circle 192 b | carrion circle 192 c |

Both B and C have lighter inner parts, but the inner gum's hue in C was pulled +5°, so it feels closer to the original reddish Mulberry Wood.

I'd go for C but if You feel it should be even brighter, I'll do some more tweaking - maybe lighten the base?
(though that might decrease the contrast between it and the gum)

palob:
C it shall be :-)

palob (approved):

PR #5551, opened 2020-08-07
Icons for Serena

maxtron95:
https://store.steampowered.com/app/272060/

serena-ci
serena-sq


palob (changes requested):
db entry needs to go after ``ser-player``

Could you please reduce the symbol height to 26px or 24px?

The photo looks a bit much yellowed, contrast needs to be beefed up significantly.

I don't think the hands are recognisable. Does the design without the hands?

palob (changes requested):
Contrast inside the photo could be livelier still (while retaining the pastel colour impression).

PR #5557, opened 2020-08-12
Icons for Popsicle

palob:
https://github.com/pop-os/popsicle

com system76 Popsicle
com system76 Popsicle-ci
com system76 Popsicle-sq

### DIFF
--- a/data.json
+++ b/data.json
@@ -16343,6 +16343,14 @@
             ]
         }
     },
+    "popsicle": {
+        "linux": {
+            "root": "popsicle",
+            "symlinks": [
+                "com.system76.Popsicle"
+            ]
+        }
+    },
     "portal": {
         "linux": {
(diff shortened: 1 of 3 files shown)


maxtron95:
@palob How does it look on a pink baseplate?

palob:
Probably worth a try, will do when I find time.

palob:
popsicle1
popsicle2
popsicle3
popsicle4

palob:
I think I prefer the purplish no 3. Could be a tad brighter though.

maxtron95:
I'd say either No. 3 or No. 4 (I like the color of No. 4 in my opinion.)

Foggalong (approved):
Quite like the original blue so running with it! Nice suggestion though @creepertron95, the pinks were still nice :)

PR #5558, opened 2020-08-13
Update Atom designs

palob:
atom-nightly
atom-nightly-ci
atom-nightly-sq


maxtron95:
@palob Should we redo the colors of the other 2 atom icons as well?

palob:
Were the colours changed upstream?
If so we should redo the symbol alongside, it isn't executed very well.

maxtron95:
@palob The Atom icons were always those colors but the color in Numix's Atom icons is white and the stable Atom icon is purple for some reason.

Atom -


Atom Beta -


Atom Dev (We should add an icon called "atom-dev") - 


Atom Nightly -

maxtron95:
@palob Should we modify the colors of the other atom icons?

palob:
Yep, I'd say so.

Foggalong (changes requested):
Agreed, let's take the opportunity to recolour them all! You up for doing that @palob? No worries if not, I'd be fine to handle it :slightly_smiling_face:

palob:
Generally speaking yes, might take some time until I can get to this as this is something I envision to take extra time.

PR #5559, opened 2020-08-14
Icons for DownZemAll

maxtron95:
https://github.com/setvisible/DownZemAll

downzemall

downzemall-ci
downzemall-sq


palob (changes requested):
This needs some fiddling with the nodes for alignment and smothness.
Did something but not finished yet

arrows

The 2 arrow parts should contrast more strongly and the upper one should cast a shadow on the lower one.

palob (changes requested):
Thank you. The hole in the middle could be a bit bigger as well as clearance between tip and tail.

Could you please increase the contrast between tip and tail. Maybe there's a good combination in the original icon.

PR #5560, opened 2020-08-14
Logo for AcademiX GNU/Linux

maxtron95:
https://academixproject.com/en/home/

dist-academix-ci
dist-academix-sq


palob (commented):
Could we please try a much lighter baseplate for comparison?

Foggalong (changes requested):
No on green, think this is gonna have to be white or dark/navy blue to avoid any colour clashes

PR #5561, opened 2020-08-14
Logo for ALT Linux

maxtron95:
https://en.altlinux.org/Main_Page

dist-alt-ci
dist-alt-sq


palob (changes requested):
This is lovely but could we get the baby penguin a bit more rounded such that it looks more akin to its parents? Maybe a change as little as this will do it.
rounded

palob:
Circle symbol should go 1px left and up.

palob (approved):

PR #5562, opened 2020-08-14
Logo for ArchLabs

maxtron95:
https://archlabslinux.com/

alt text

dist-archlabs-ci
dist-archlabs-sq


palob (changes requested):
There's a few alignment issues with the flame, the stroke around it should be (roughly) 1px wide.

palob (approved):
Will merge this right away, didn't find a way to both improve alignment and keep proportional characteristics.

PR #5565, opened 2020-08-18
Dedicated icons for 1Password

maxtron95:
For #5379 

https://discussions.agilebits.com/discussion/114964/1password-for-linux-development-preview

1password-ci
1password-sq


Foggalong (changes requested):
I've come back to this one a couple of times and it still looks weird to me. Can we get it slightly smaller and on a 30px white circle a la the original so that it's more clear it's a key hole?

palob:
In addition the corners should probably be rounded like in the original.

mrmeszaros:
Hey, I did rework this slightly - what do You think?

Original | Cirlce | Square
-- | -- | --
1password original 192 | 1password circle 192 | 1password square 192

Though I am not totally sure about the drop shadow - maybe it should "indent" like it's a hole?

palob:
I'd say "indent" shadow on the keyhole and standard drop shadow for the white circle.

PR #5567, opened 2020-08-21
Icons for Azure Data Studio

maxtron95:
https://docs.microsoft.com/en-us/sql/azure-data-studio

azuredatastudio-ci
azuredatastudio-sq


palob (changes requested):
Could you please try a different baseplate colour which contrasts better to the cog?
Maybe light blue, light green, cyan?

palob:
Also wondering how a shadow for the wavelet would look like.

Foggalong (changes requested):
Rather than the baseplate, make the cog mid-light grey, it'll keep the nice contrasts elsewhere. I thiiink the wavelet shadow might be a bit much though

PR #5568, opened 2020-08-21
Icons for ILSpy 

maxtron95:
https://github.com/icsharpcode/AvaloniaILSpy

ilspy-ci
ilspy-sq


palob (changes requested):
very minor itch: The highlight on the magnifying glass should be a quarter circle for optimal alignment.

PR #5574, opened 2020-08-25
Icons for Barotrauma

maxtron95:
https://store.steampowered.com/app/602960

barotrauma-ci
barotrauma-sq


palob (changes requested):
There should be a shadow on the path beneath the boat.
Could you please make the tail fin a bit smaller and smoothen the boat front a bit?

PR #5575, opened 2020-08-25
Icons for Bashtop and Bpytop

maxtron95:
https://github.com/aristocratos/bashtop

bashtop

bashtop-ci
bashtop-sq

https://github.com/aristocratos/bpytop

bpytop-ci
bpytop-sq


palob (changes requested):
Both symbols should be slightly brighter in the bottom part for the sake of better readability.

palob:
@creepertron95 Just asking, are you OK with me finishing PR with only trivial changes needed like this one?

PR #5579, opened 2020-08-26
Icons for bauh

maxtron95:
https://github.com/vinifmor/bauh

bauh-ci
bauh-sq

### DIFF
--- a/data.json
+++ b/data.json
@@ -1545,6 +1545,11 @@
             "root": "steam_icon_238460"
         }
     },
+    "bauh": {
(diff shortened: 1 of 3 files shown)


palob (changes requested):
I really don't like those coloured spots from the original icon. I'm sure the icon can do without.
Could you please make the bottom edges less prominent (vaguely reminiscent of a hat right now), more like in the original icon?

palob (approved):
Thank you for coming back!

PR #4048, opened 2017-12-23
Use Docker with Travis

bilelmoussaoui:
See #4047 
This is not ready yet.

PR #5587, opened 2020-09-11
Specific icon for OpenERP

Foggalong:
Actually just a recolour of Opera's original icon, but who's counting


Foggalong:
Hmmm, seems there's some significant cruft being left in the circle icon. Will investigate what that is.

Foggalong:
Just an extra gradient, all good! :+1: Gonna take a look at the other Opera icons too and update them to the new design

Foggalong:
Also a cheeky reuse of the old Opera Next colour scheme for the refreshed Opera Widget icon :grin:

Foggalong:
@palob From what I can tell there aren't actually any build issues, but Travis isn't happy that our tester depends on Python 3.4 - pinged Bilal about sorting that in #4048

palob (approved):

palob:
As an aside, OpenERP seems to be [odoo](https://www.odoo.com) these days.

Foggalong:
Oh my word, if I'm reading this right what was one app has since been split into over 40. If someone requests we'll cover it but probably not one to go out of our way to preempt.

PR #5588, opened 2020-09-15
Symlinks aplenty

palob:

### DIFF
--- a/data.json
+++ b/data.json
@@ -7612,7 +7612,10 @@
     },
     "fontmatrix": {
         "linux": {
-            "root": "fontmatrix"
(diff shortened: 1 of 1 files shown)


Foggalong:
I've been meaning to ask: is there a technical reason you're using a different name for the symlink branch each time?

palob:
It's just a habit to avoid the clash in case the old one was still existing.

Foggalong:
That's totally okay, I just wanted to check I wasn't causing any unnecessary problems 😅

Foggalong (approved):

PR #5600, opened 2020-10-05
perspective fix for help-browser

palob:
The current design looks cool but doesn't makes sense perspective-wise and isn't used elsewhere.


Foggalong (approved):
Huh yeah, I've never thought about it but this icon implies a 3D perspective unlike basically every other icon? Good catch!

PR #5622, opened 2020-11-07
Update Inkscape usage for new versions of inkscape

Supernova4422:
If inkscape is installed with Brew, the current commands will fail. This will ensure that inkscape works. Worth noting that we should use subprocess.run so error checking can be done.


Foggalong:
From what I can the old code still works with the latest version of Inkscape on Linux, but changing it doesn't seem to affect anything for Linux users so if it fixes things for macOS users I'm happy to merge :+1:

Foggalong (approved):
From

PR #5624, opened 2020-11-10
Xfce symlinks

palob:
Fixes #5623

PR #5684, opened 2021-02-12
add a few xfce entries

bgstack15:
In tandem with #5624, here are a few extra xfce entries for xfce4-4.16.0 from Debian.

You will note that this patch does not include the "org.xfce.terminal" that 5624 does, but I don't know how to submit a patch onto that pull request.


palob:
Hi, thank you for the PR.
``org.xfce.terminal-settings`` should be reordered, this is why the CI build fails.

This PR should not add any icon names which are already present in #5684, this way both PRs can be merged in sequence without further ado.

bgstack15:
I have fixed the order for org.xfce.terminal-settings as indicated.

And I assume you meant that I should make sure I do not duplicate any new entries from #5624. I had already reviewed it before submitting this initially, and these contents do not overlap but some of the additions are adjacent so unfortunately some manual fixing of the patches might be required.

palob (approved):

bgstack15:
Does this need anything from my side to proceed?

palob:
No, you've done your share.

PR #5690, opened 2021-02-14
alignment fixes for knights and stuntrally

palob:


Foggalong (approved):
LIke your change for stuntrally! Much more Numixy

PR #5726, opened 2021-04-15
Icons for Friday Night Funkin'

maxtron95:
https://github.com/ninjamuffin99/Funkin

funkin-ci
funkin


palob:
Is it a fish or a guy wearing a hat? Is there a way to have the 3 white shapes look like eyes + mouth?
Maybe the eyebrow line connecting the eyes + the line left to the mouth is needed?

maxtron95:
@palob The icon is supposed to be the main character, the boyfriend from the game.

funkin-icon

palob:
Ah, I see, I'd say just use those colours sans the outlines, will be much more recognisable.

maxtron95:
I think the baseplate would have to changed as well because I think the new colors would clash with the current baseplate color.

Foggalong (changes requested):
Agree with both, add skin shape and change b/g colour - a dark purple might work?

PR #5735, opened 2021-05-23
Updated qualculate application icons to stay closer to original icon …

gruenfron:
| Original icon |
| --- |
|   | 

|  | Circle | Square |
| --- | --- | --- |
| Current |  |  |
| Updated |  |  |

Description: The current Qalculate! circle and square icons are IMHO a bit far away from the original design.


palob (changes requested):
Hi, thank you for your contribution. There are a few issues which need to be adressed:

+ All elements, especially horizontal, vertical and circular ones need to be aligned with the pixel grid (to be toggled with the ``#`` key in Inkscape).

The **=** for instance currently looks like this

misalign

It should look like this

misalign2

+ Right now you are using font glyphs for the X and the omega. Those need to be converted to paths or else the icon will be broken for using not having the font installed.

+ Please do the symbol shadows manually by duplicating (and unifying) the elements. Shadow colour is full black ``000000ff`` with 10% opacity. Shadow displacement is 1px down for Square and 1px down + 1px to the right for Circle

+ Your icon has elements exceeding the baseplate

+ The baseplate is misplaced and misshaped. It should be 46px X 46px in size, placed at X=1px, Y=1px. Basing the icon on the templates ``48.svg`` in the ``templates`` folder should fix this.

+ Please don't group elements because this will increase file size.

+ X and omega should fit into a 12px X 12px rectangle like the symbols in the ``calc.svg`` icon

12x12

or like this omega sketch

omega

+ The right part should be coloured with a linear gradient, 4L (10L for Inscape pre 1.0) lighter at the top stop. Please open existing icon to get an impression of how it's done.

+ Please save the file as *Optimised SVG* with options as shown here
https://github.com/numixproject/numix-core/wiki/Optimise-Options

Don't feel intimidated by the number of issues. You can solve those step by step and ask questions any time.

palob:
I've just fixed a few small issues with our ``calc.svg`` in the main branch, so you PR is paying off twofold :-)

gruenfron:
Hi and thanks for the detailed issue description, this helped me a lot as a total Inkscape noob. The fixes hopefully cover the issues so far.

|  |   |
| --- | --- |

There is one thing I made a bit differently compared to `square calc.svg`. The symbols (χ,=,Ω) for the square icon are more centered in their quadrants, so the size of the 'bounding box' is larger than maybe wanted (26x26 as in calc, or 24x24 as in template). The center-narrowd symbol location in `square calc.svg` or `square multisystem.svg` look a bit unbalanced to my eye :grin:.

palob (changes requested):
Sorry for taking so long for the next review, had a busy time.

Thank you for the changes. There are a few remaining issues which require some work but no "SVG design magic" I think.
+ In the Circle icon there is something wrong with the stacking order of the baseplate parts.
The 2 lower parts need to be moved up to be atop of the baseplate bezel path. Right now the bezel looks too dark because of this.
+ The Ω is slightly misplaced/wobbly. It should be symmetrical (horizontally), the nodes reveal it isn't
omega
You can achieve this by subtracting a smaller ellipse from a bigger one and cutting of the lower part of the resulting ring, e.g. like this
omega2
+ Ω and x (or is it the greek literal χ?) still look a bit big, height should be 10 px at max for both. The **=** sign should get a tighter spacing with only 2 px between the line.
+ Please remove the gradient shadow on top of the right icon half. This is a feature not to be found in any of the existing icons.
+ The 3 baseplate parts need to overlap in order to avoid fine outlines due to a bug? present in SVG renderers (librsvg, QtSVG). For instance like shown here (transparency added for illustration purposes)
overlap
+ The vertical baseplate gradient should only alter the L value, hue and saturation should remain constant throughout the gradient.
+ Please use ``f9f9f9ff`` (*2,5% Grey* in Inkscape) instead of full-blown white.

palob:
@gruenfron Will you finish this? If not I can step in in the next weeks.

palob:
Took this over.

PR #5753, opened 2021-09-19
Use Github Actions instead

barelyhuman:
For:  https://github.com/orgs/numixproject/teams/team/discussions/2?from_comment=1#discussion-2-comment-1


satya164 (commented):
LGTM but I think it'd be good to have some way of caching in future to avoid installing these deps every time

Foggalong:
@satya164 We have talked about using a Docker instance to handle that for Travis. We got as far as setting up a Numix account + instance before the contributor's (@bilelmoussaoui) time dried up

See #4048 for more details

barelyhuman:
@satya164 can be cached if I move them to a requirements file and use pip cache instead, I’ll add that early tomorrow to the same

barelyhuman:
@Foggalong done, 

the time reduced from 1 min 5 seconds to 35-45 seconds that is considering there's no updates needed and only installing the needed packages from the pip cache and checking if they need an upgrade, we can further reduce this by freezing the versions of each package

.github/workflows/test.yml
```diff
@@ -0,0 +1,38 @@
...
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.6
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+
+      - uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ hashFiles('test-requirements.txt') }}
```
Foggalong: This is looking great! Very minor change request, can we keep `test-requirements.txt` in the same directory as the workflows rather than at root

Foggalong (changes requested):

.github/workflows/test.yml
barelyhuman: Can do

barelyhuman: Done

Foggalong (approved):

Foggalong:
Will merge this tonight and clean up any Travis cruft left on our account :+1:

Foggalong:
@barelyhuman Was there a particular reason for using `--upgrade --upgrade-strategy eager`? I was tinkering with the actions today because I'm adding a new step to fix #4786 and noticed use of those flags constitutes _half_ of the builds runtime. I've removed them in `python-clean` where I'm working but if there was a reason let me know and I'll add it back :slightly_smiling_face:

barelyhuman:
Thats in case we froze the versions, since we didn’t it’s fine

Foggalong:
Aces, cheers @barelyhuman!

PR #5757, opened 2021-10-13
Icons for Lovers in a Dangerous Spacetime

### LINKED ISSUE #5465: Icon for Lovers in a Dangerous Spacetime
| --- | :-- | :-- | :-- |
| `Lovers in a Dangerous Spacetime`| `steam_icon_252110` | `Lovers in a Dangerous Spacetime.desktop` | steam_icon_252110 |

Description: Cute co-op 2D spaceship shooter

Website: https://www.loversinadangerousspacetime.com/

really-need-an-api-key:
Long overdue. Fixes #5465.

| --- | :-- | :-- |
|  image |  circle-lovers-in-a-dangerous-spacetime | square-lovers-in-a-dangerous-spacetime |


palob (changes requested):
+ For Circle the symbol can go down one more px
+ The face of the right guy should be brighter, at least L=18
+ The circle on top of the right guy's helmet should be of 2px diameter so it can be placed at integer coordinates

really-need-an-api-key:
Doesn't the antenna look a bit too squarish at 2px?
image image
And looking at it now I think have to make yellow guy's eyes a bit brighter when the face is at L=18

palob:
It looks more squarish but crisper, even more so when drawn at 24px X 24px (which fortunately isn't done that frequently anymore in today's DEs). The problem is that the antenna shaft is placed at .5 px X coordinate so it looks just as wide but that can't be helped probably.
Eyes can be brighter, yes.

really-need-an-api-key:
Should I also align the hemlet to the shoulder like so?
image

palob:
No, it's fine as is with the ring properly aligned vertically.

Could you please extend the paths like this to avoid shine-through artifacts (transparency for illustration)?

extend

really-need-an-api-key:
I think the square one could also be moved down 1px. Or maybe just the heart shape, while leaving the rest intact, but that change should also be done in the circle version then.
L is original, M is everything -1px, R is heart -1px.
a b c

Which one looks more centered to you?

palob (changes requested):
Thank you for the tweaks. The Square icon has the shadows misplaced horizontally.

really-need-an-api-key:
I really have to be more mindful about the arrow keys moving things in 2px increments in Inkscape...

Oh, I guess I also forgot to move the 2nd shadow. Will fix that in a second.

palob:
You can change that in Preferences section ``Behaviour -> Steps``.

palob (approved):

PR #5761, opened 2021-10-16
Icons for Microsoft Edge

### LINKED ISSUE #5751: Microsoft Edge
- **Name:** Microsoft Edge
- **Description:**: Web browser by Microsoft
- **Website:** https://www.microsoftedgeinsider.com/en-us/
- **Icon Name:** `ms-edge`
- **Symlinks:** `Edge`,`microsoft-edge`, `microsoft-edge-beta`,`com.microsoft.Edge`
- **Desktop File:** `microsoft-edge-beta.desktop`
- **Original Icon:**

Credit to [Arjun](mailto:arjun6280414@gmail.com) for reporting this issue via email.

really-need-an-api-key:
Tried to keep it as aligned as possible.
Fixes #5751.

| --- | :-- | :-- |
| image | circle-ms-edge | square-ms-edge


palob:
Could you please try a more neutral (less saturated background), splitting the upper path into two parts for a smoother colour transition and adding shadows between the paths similar to the original icon?

really-need-an-api-key:
Not really sure how to split the upper part, so ended up doing this.
I think two parts would make for a too noticeable transition.
ms-edge
Looks better with 3 imo.
ms-edge-2
What do you reckon?

palob:
Yeah, splitting into 3 looks ace :+1:

really-need-an-api-key:
circle-ms-edge
square-ms-edge

palob:
Hi, can you please fix the alignment to something like this (facilitating new nodes).
This is just a quick and dirty sketch, paths below need tweaking as well, will take some fiddling.

alignment

Now with the new seperate paths there should be shadows for each.
Is it possible to keep the colourful impression of the original icon by colouring the tip according to the original icon?

Is it possible to have both baseplate and symbol brighter still?
While redesigning much of our existing icons I stumbled upon the fact that we've got too many rather dark icons so brighter ones are welcome.

really-need-an-api-key:
I've not changed the colours yet, but by shadows for each part do you mean in the marked areas?
image

Because I was thinking the three "gradient" parts are more of a "virtual" division rather than distinct parts, as there is no shadow cast by the green part onto itself in the original icon.

palob:
Yep, it's those marked areas I had in mind.
I think having differently coloured sections with and w/o shadows in between looks strange.

Now maybe there is a way to go back to the initial 3-colour icon by moving the colours closer together so there's no stark contrast to the bigger front part?
Maybe ditching the vivid green or the dark blue?

palob:
Twiddled a bit and this came out:
edge

Sorry for being that late.
@really-need-an-api-key Are you OK with me pushing this and merging?

really-need-an-api-key:
Sure

PR #5796, opened 2022-09-22
Add icon for Dark Echo

mrmeszaros:
| Original icon | Circle icon | Square icon | Banner |
| --- | :-- | :-- | --- |
| dark-echo original | dark-echo circle 192 | dark-echo square 192 | image |

A game about navigating through the dark.


palob (commented):
There's an ever so slight issue at the gradient. Stop color at the cottom should be ``#303030`` (L24) so the gradient spans from L20 to L24. 
By the way, are you perhaps using an older Inkscape version where the L range spans from 0 to 360 (instead of 0 to 100) so this might be a conversion error?

mrmeszaros:
I am a bit confused: the gradient goes from `#3d3d3d` to `#303030` (top to bottom).
Inkscape rounds `#303030` to L19.
I did a custom calculation: `#303030` is L18.82 says L20 would be `#333333` exactly.
Should I change it?

BTW my Inkscape is at: `Inkscape 1.2 (1:1.2.1+202207142221+cd75a1ee6d)`.

palob:
I've just checked this and we've got both in existing icons: ``#3d3d3d`` to ``#303030`` and ``#3d3d3d`` to ``#333333``.

The latter seems to be what was carried over from icons desgined according to our guidelines of a 10L gradient for the old 360 lightness scaling.

So I'd say let's change it to ``#333333``. I'll open a ticket to tackle this (low-priority) issue some time.

palob (approved):

PR #5798, opened 2022-09-30
Add icon for Moon Hunters

mrmeszaros:
Fixes #5427

| Banner | Circle icon | Square icon |
| --- | --- | --- |
| moon-hunters base | moon-hunters circle 384 | moon-hunters square 384 |

Colors seem a bit too bright, currently working on darker shades.
Also, am not fully satisfied with the hair - looking into alternate lining + maybe including that circle hair ornament.

PR #5803, opened 2022-10-09
Fix minor Moon Hunters icon issues

mrmeszaros:
This will fix some minor issues with the icon that were left out of PR #5798.

* better align hair line to the grid
* update the drop shadow to reflect the symbol outline changes
* remove unreferenced ids


palob (approved):
Oh, there was even more that slipped me. Thank you for the followup.

PR #5815, opened 2022-11-09
Add icon for 1password

mrmeszaros:
Original icon | Circle icon | Square icon
-- | -- | --
1password original 192 | 1password circle 192 | 1password square 192

This is a rework of, and so closes #5565.


palob (approved):
Thanks again.

PR #5889, opened 2025-11-29
Icons for Dicomizer & symlink Weasis → weasis

Merikei:
| --- | :-- | :-- |

The same as [weasis.svg](https://github.com/numixproject/numix-core/blob/master/icons/square/48/weasis.svg) except for the background


palob (approved):
Thank you.

PR #5896, opened 2025-12-31
org.jellyfin.JellyfinDesktop → jellyfin

Merikei:
https://github.com/jellyfin/jellyfin-desktop/blob/771fcb73039f0235ec9aa9f1a010c74570330262/resources/meta/org.jellyfin.JellyfinDesktop.desktop#L6


palob (approved):
Thanks.

PR #5900, opened 2026-04-17
Add FocusTimer to Linux symlinks

BlueManCZ:
gnome-pomodoro changed its icon name, reflecting the change as symlink.


BlueManCZ:
Just noticed it's already covered by `pomodorolm`. My bad, closing.

